PR #1, opened 2018-04-19
Add basic navigation support

MattCheely:
This is still pretty rough around the edges, but at least it enables basic link navigation from components. Direct url navigation was implemented in the initial code used to set up the repo, which I didn't PR, although I probably should have, in retrospect.

I'm going to set up an epic for this work soon, but next steps after this are:

- Clean up package.json & publish
- Add support for exchanging auth token with parent
- Initial testing in web-dir branch with a bunch of clean-up in parallel

@justinaray @jrrucker @JackieChiles @omarestrella @bjdupuis


src/elm/ComponentHelper.elm
```diff
@@ -53,19 +72,33 @@ update msg model =
...
-                [ Decode.map Navigate Navigate.decoder ]
-            )
-        )
-        value
-        |> Result.withDefault (Unknown value)
+port componentIn : (Decode.Value -> msg) -> Sub msg
 
 
-port parentMessage : (Decode.Value -> msg) -> Sub msg
+port componentOut : Decode.Value -> Cmd msg
```
MattCheely: I'm starting to settle on a structure for message passing where there's a pair of input/output ports for talking to each independent app. So the component helper has a pair of ports for sending and receiving messages with the component app, and a pair of ports for sending and receiving messages with the coordinator (parent window)

src/elm/Message/LabeledMsg.elm
```diff
@@ -0,0 +1,40 @@
+module Message.LabeledMsg exposing (decoder, encode)
```
MattCheely: This module provides basic support for a standard message format between Elm & JS/TS:

```javascript
{
  msgType: "MessageType",
  msg: { ... message data here ...}
}
```

PR #3, opened 2018-05-01
Fix issue in Chrome. Build debug versions and production versions.

MattCheely:
The custom element API spec doesn't allow modification of the DOM in the element constructor. I missed that detail, and the polyfill can't enforce that check, so things were blowing up in chrome. This fixes that issue.

Additionally, the build scripts are further improved to generate js files with the Elm debugger enabled, which are helpful for testing when embedding in other apps. 

@xdumaine @justinaray @jrrucker @JackieChiles @omarestrella @bjdupuis

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,8 @@ elm-stuff
 # elm-repl generated files
 repl-temp-*
 
-# Dependency directories
+# Dependency directories & files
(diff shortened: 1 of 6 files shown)


MattCheely:
Whoops. Thanks for catching that, @xdumaine .

PR #4, opened 2018-06-01
Updates to support admin embedding demo

MattCheely:
I made a number of updates to the project to support a demo embedding a custom build of the admin UI.

It's a bit of an odd merge point for a public repo, since it's configured to demo some private code, but most of these changes are pushing in the right direction so I'd like to go ahead and get it merged. 

The next step is a bigger re-organization to split the demo and the various logical libraries into separate pieces, and I'd like to keep code improvements separate from that.

@xdumaine @justinaray @jrrucker @JackieChiles @omarestrella @bjdupuis


README.md
```diff
@@ -4,10 +4,16 @@ Utilities for composing multiple web-apps into a single application interface vi
 
 ## Setup
 
-`yarn install`
+`npm install`
```
MattCheely: Hot take

src/elements/component-frame.js
```diff
@@ -9,14 +9,14 @@ export default class ComponentFrame extends HTMLElement {
 
     constructor() {
         super();
-
         this.iframe = document.createElement('iframe');
         this.iframe.setAttribute('frameborder', 0);
         this.iframe.setAttribute('style', IFRAME_STYLE);
-        this.iframe.setAttribute('sandbox', 'allow-scripts');
+        this.iframe.setAttribute('sandbox', 'allow-scripts allow-same-origin');
```
MattCheely: this `allow-same-origin` was so I wouldn't have to jump through a bunch of CORS hoops right now. I'm not sure I want to leave it long term, as combined with `allow-scripts` it would allow components to break sandboxing.  But without out it, every network request is sent with a `null` origin, which makes every request cross-origin.

README.md
jrrucker: o_O

jrrucker:
I don't have an approve button or anything, but LGTM

MattCheely:
Fixed during code reorganization.

PR #9, opened 2018-09-13
Basic typescript conversion.

MattCheely:
This is the bare minimum to get things working with typescript. I'll open a follow-up PR with improved type definitions and such, but I wanted to get this in place first.


js/iframe-coordinator/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "outDir": "./",
+    //"noImplicitAny": true,
```
MattCheely: I will definitely come back and re-enable this on the second path to make sure everything is well typed.

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -0,0 +1,28 @@
...
+
+class FrameRouterElement extends HTMLElement {
+  router: HostProgram;
+
+  constructor() {
+    super();
+    this.setAttribute("style", "position: relative;");
+  }
+
+  registerComponents(components) {
```
rbrcurtis: you should add return types.

PR #10, opened 2018-09-13
Refactor with updated host/client terminology.

MattCheely:
The terms "coordinator" and "component" for the top-level and embedded
apps were a bit confusing. An informal survey of developers indicated
that host and client were significantly less confusing.


src/ClientMessage.elm
```diff
@@ -1,54 +1,54 @@
...
+@docs ClientMessage, decoder, encode
 
 -}
 
 import Dict exposing (Dict)
 import Json.Decode as Decode exposing (Decoder)
 import Json.Decode.Pipeline exposing (decode, required)
 import Json.Encode as Encode
-import Message.LabeledMsg as LabeledMsg
+import LabeledMessage
```
justinaray: As I was working on toast message, I added the type, encoders, and decoders in Message.ToastMsg.  It might be nice to have this root message module for code organization as the message types grow.  That said, I'm still an elm padwan and not sure if this is the best way to organize.

justinaray: Along the same lines, maybe it would be good to have top level modules of host, client, and message?

Again, take this with a grain of salt.

MattCheely: For ToastMsg, since it's a client -> host message (ignore any action responses for now), I'd add all of the decoders and whatnot in this file directly, and add a variant to the `ClientMessage` type to capture it.

A good organizational strategy in Elm is to keep more code than you normally would in JavaScript in a single file and only split when things start to get awkward. Having a good type system makes putting more code in one place less of an anti-pattern. (there are limits, of course)

PR #11, opened 2018-09-14
Move all DOM work out of custom element constructors.

MattCheely:
My takeaway here: Never do _anything_ in a custom element constructor.

### DIFF
--- a/js/iframe-coordinator/elements/x-ifc-frame.ts
+++ b/js/iframe-coordinator/elements/x-ifc-frame.ts
@@ -12,34 +12,41 @@ class ClientFrame extends HTMLElement {
 
   constructor() {
     super();
+  }
+
+  connectedCallback() {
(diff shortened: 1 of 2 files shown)


MattCheely:
I won't merge this until you can confirm things are working, @rbrcurtis

js/iframe-coordinator/elements/x-ifc-frame.ts
```diff
@@ -12,34 +12,41 @@ class ClientFrame extends HTMLElement {
 
   constructor() {
     super();
+  }
+
+  connectedCallback() {
```
justinaray: Wow!  I could see appending, but not even standard DOM apis?  Interesting and a good lesson learned.

rbrcurtis:
> I won't merge this until you can confirm things are working, @rbrcurtis

it works!

PR #12, opened 2018-09-14
Fast-fail buildall script

justinaray:
Set -e option to fail the build on elm compilation errors.  Otherwise, the demo app starts unconditionally.


scripts/buildall.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
```
justinaray: necessitated by my inability to write elm code that actually compiles ;)

PR #13, opened 2018-09-18
Pcui 403 pub sub

MattCheely:
Still a work in progress, but I wanted to get it out for early review. This gets publication from client to server working. I will add some docs soon.


js/demo-app/src/component.js
```diff
@@ -1,3 +1,19 @@
...
+// Start intercepting link click events for routing
+iframeClient.start();
+
+document.getElementById("do-publish").addEventListener("click", () => {
+  //Get the topic and the data to publish from the form
+  let topic = document.getElementById("pub-topic").value;
+  let data = JSON.parse(document.getElementById("pub-data").value);
+
+  //Publish the data
+  iframeClient.publish(topic, data);
```
MattCheely: Publishing is pretty straightforward from the client.

js/demo-app/src/index.js
```diff
@@ -14,3 +17,15 @@ document.getElementById("router").registerClients({
...
+router.subscribe("publish.topic");
+
+// Listen to publication events (will only be emitted for subscribed topics)
+router.addEventListener("publish", event => {
+  let publication = event.detail;
+  console.log(
+    `Recieved pub-sub data on topic ${publication.topic}:`,
+    publication.payload
+  );
+});
```
MattCheely: I'm not sure this is the best API on the host side, but it works for now.  Subscribing to publish events from the `frame-router` element will watch all pub-sub events, but events will only be emitted for topics the host has subscribed to via `subscribe`

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -21,9 +33,18 @@ class FrameRouterElement extends HTMLElement {
 
   registerClients(clients) {
     this.router = Host.embed(this, clients);
+    this.router.ports.toHost.subscribe(message => {
+      if (message.msgType == "publish") {
+        let event = new CustomEvent("publish", { detail: message.msg });
+        this.dispatchEvent(event);
+      }
+    });
```
MattCheely: When the host library receives a publication from the client, it checks within Elm to see if the host is subscribed, if so, the event is emitted on the `toHost` port and dispatched as a custom DOM event here.

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -21,9 +33,18 @@ class FrameRouterElement extends HTMLElement {
...
+  }
 
-    window.addEventListener("message", event => {
-      this.router.ports.fromClient.send(event.data);
+  subscribe(topic: string): void {
+    this.router.ports.fromHost.send({
+      msgType: "subscribe",
+      msg: topic
     });
```
MattCheely: When the host app subscribes, we send a message to the Elm code on `fromHost` to trigger the internal subscription state change.

js/iframe-coordinator/webpack.config.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
         exclude: [/elm-stuff/, /node_modules/],
         use: {
           loader: "elm-webpack-loader",
-          options: {}
+          options: { debug: true }
```
MattCheely: I need to fix this so it's not always enabled.

src/ClientMessage.elm
```diff
@@ -81,3 +96,12 @@ encodeLocation loc =
...
         ]
+
+
+
+-- Publish
+
+
+publishDecoder : Decoder ClientMessage
+publishDecoder =
+    Decode.map Publish CommonMessages.publicationDecoder
```
MattCheely: For folks not familiar with decoders, this is describing the process of "Decode data with `CommonMessages.publicationDecoder` (which decodes to Publication) and then use the `Publish` constructor to tag it as a variant of `ClientMessage`"

src/HostProgram.elm
```diff
@@ -1,38 +1,46 @@
...
+port binding is handled in the custom frame-router element in LINK\_TO\_JS\_LIB\_HERE
 -}
-create : ((Decode.Value -> Msg) -> Sub Msg) -> Program Decode.Value Model Msg
-create inputPort =
+create :
+    { fromHost : (Decode.Value -> Msg) -> Sub Msg
+    , toHost : Decode.Value -> Cmd Msg
+    }
+    -> Program Decode.Value Model Msg
+create ports =
```
MattCheely: The HostProgram now has more ways it needs to send/recieve data from JavaScript, so the port arguments it gets on creation changes.  This wouldn't be needed if this wasn't designed to eventually be published on the Elm package system to take advantage of enforced SemVer when we change the API.

src/HostProgram.elm
```diff
@@ -125,19 +169,29 @@ src value =
     attribute "src" value
 
 
+onClientMessage : Attribute Msg
+onClientMessage =
+    on "clientMessage" (Decode.field "detail" ClientMessage.decoder |> Decode.map ClientMsg)
```
MattCheely: When there's a `clientMessage` event, pull the data from the `detail` field on the event, decode it to a `ClientMessage` and tag it as `ClientMsg` to convert it to this program's `Msg` type, for handling in `update`

PR #14, opened 2018-09-19
Add toast message support

justinaray:
toastmsgdemo

### DIFF
--- a/js/demo-app/public/components/example1/index.html
+++ b/js/demo-app/public/components/example1/index.html
@@ -8,12 +8,14 @@
     <li><a href="#/two/second">/two/second</a></li>
 </ul>
 
+<button class="toast" data-component-name="Component #1">Send Toast</button>
+
 <script type="text/javascript" src="/component-app.js"></script>
 <script type="text/javascript">
- console.log("RENDERED");
- document.getElementById("path").innerHTML = window.location.hash;
-
- window.onhashchange = function () {
-    document.getElementById("path").innerHTML = window.location.hash;
- }
+  console.log("RENDERED");
+  document.getElementById("path").innerHTML = window.location.hash;
+  
+  window.onhashchange = function () {
+      document.getElementById("path").innerHTML = window.location.hash;
+  }
 </script>
--- a/js/demo-app/src/component.js
+++ b/js/demo-app/src/component.js
@@ -1,3 +1,21 @@
 import component from "iframe-coordinator/client.js";
 
-component.start();
+let worker = component.start();
+
+const TOAST_LEVELS = ['info', 'success', 'error']
+
+document.addEventListener('DOMContentLoaded', () => {
+  let toastBtnEl = document.querySelector('button.toast');
+  toastBtnEl.addEventListener('click', () => {
+    let options = {
+      title: 'Hello iframe World',
+      // Custom, App-specific props here
+      custom: {
+        level: TOAST_LEVELS[Math.round(Math.random() * 2)]
+      }
+    };
+
+    worker.requestToast(`From ${toastBtnEl.getAttribute('data-component-name')}`, options);
+  });
+});
+
--- a/js/iframe-coordinator/elements/frame-router.ts
+++ b/js/iframe-coordinator/elements/frame-router.ts
@@ -1,13 +1,30 @@
 import { Host } from "../elm/Host.elm";
 
+const TOAST_REQUEST_MSG_TYPE = 'toastRequest';
+const TOAST_REQUEST_EVENT_TYPE = 'toastRequest';
+
 interface Host {
   ports: {
     fromClient: {
       send: (message) => void;
     };
+    toHost: {
+      subscribe: (callback: (message) => void) => void;
+    };
   };
 }
 
+/**
+ * The frame-router custom element
+ * 
+ * Events:
+ * @event toastRequest
+ * @type {object} 
+ * @param {object} detail - Details of the toast.
+ * @param {string} detail.message - Toast message.
+ * @param {string=} detail.title - Optional toast title.
+ * @param {object=} detail.x - Optional, custom properties for application-specific toast features
+ */
 class FrameRouterElement extends HTMLElement {
   router: Host;
 
@@ -24,6 +41,15 @@ class FrameRouterElement extends HTMLElement {
 
     window.addEventListener("message", event => {
       this.router.ports.fromClient.send(event.data);
+    });
+
+    // Subscribe to component out port to handle messages 
+    this.router.ports.toHost.subscribe(labeledMsg => {
+      if (labeledMsg.msgType === TOAST_REQUEST_MSG_TYPE) {
+        this.dispatchEvent(new CustomEvent(TOAST_REQUEST_EVENT_TYPE, {detail: labeledMsg.msg}));
+      } else {
+        console.error('Unexpected Message from Host Program', labeledMsg.msgType);
+      }
     });
   }
 }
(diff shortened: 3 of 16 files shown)


js/demo-app/public/components/example1/index.html
```diff
@@ -8,12 +8,14 @@ <h1>Component #1</h1>
     <li><a href="#/two/second">/two/second</a></li>
 </ul>
 
+<button class="toast" data-component-name="Component #1">Send Toast</button>
```
justinaray: @MattCheely I can easily add another tab to the nav and a more full-featured demo for the toasts.  It could include title, message, and level using form fields.  Thoughts?  Sufficient for now?

js/demo-app/src/component.js
```diff
@@ -1,3 +1,19 @@
 import component from "iframe-coordinator/client.js";
 
-component.start();
+let worker = component.start();
+
+const TOAST_LEVELS = ['info', 'success', 'error']
+
+document.addEventListener('DOMContentLoaded', () => {
+  let toastBtnEl = document.querySelector('button.toast');
+  toastBtnEl.addEventListener('click', () => {
```
justinaray: @MattCheely I can easily add another tab to the nav and a more full-featured demo for the toasts. It could include title, message, and level using form fields.

Just wanted to get this PR up for comment so I didn't build it out too much.  Thoughts?

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -1,13 +1,30 @@
 import { Host } from "../elm/Host.elm";
 
+const TOAST_REQUEST_MSG_TYPE = 'toastRequest';
+const TOAST_REQUEST_EVENT_TYPE = 'toastRequest';
```
justinaray: Should we export this constant so it's available for users?  The event and type is documented below.

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -25,6 +42,15 @@ class FrameRouterElement extends HTMLElement {
     window.addEventListener("message", event => {
       this.router.ports.fromClient.send(event.data);
     });
+
+    // Subscribe to component out port to handle messages 
+    this.router.ports.toHost.subscribe(labeledMsg => {
```
justinaray: I changed this to encode and send the entired LabeledMessage object so that I knew the event type.

If we ever add the `protocol` field we discussed today or other meta fields, we might have to revisit this or wrap LabeledMessage with another class internally.

Let me know though if you see issues with exposing this object out this far.

js/iframe-coordinator/elm/Host.elm
```diff
@@ -6,7 +6,10 @@ import Json.Decode as Decode
...
 main : Program Decode.Value HostProgram.Model HostProgram.Msg
 main =
-    HostProgram.create fromClient
+    HostProgram.create fromClient toHost
 
 
 port fromClient : (Decode.Value -> msg) -> Sub msg
+
+
+port toHost : Decode.Value -> Cmd msg
```
justinaray: Adds the port to communicate with the custom element

js/iframe-coordinator/libs/client.ts
```diff
@@ -23,10 +23,45 @@ function start(expectedOrigin) {
...
+     * 
+     * The page embedding the host is responsible for handling the fired custom event and
+     * presenting/styling the toast.  Application-specific concerns such as level, TTLs,
+     * ids for action callbacks (toast click, toast action buttons), etc. can be passed via custom
+     * properties of the options param.
+     * 
+     * @param {string} message - The message content of the toast
+     * @param {object=} options - Supplimental toast options.
+     * @param {string=} options.title - Optional title for the toast.
+     * @param {*=} options.x - Optional, application-specific toast properties.  Note: Must be JSON serializable.
```
justinaray: Did a fair amount of looking, but I'm not sure dynamic properties are supported in jsdoc.  Let me know if y'all know of a supported syntax.  If not, this should be pretty obvious from a textual perspective what we're after.

src/HostProgram.elm
```diff
@@ -1,34 +1,35 @@
-module HostProgram exposing (create, Model, Msg)
+module HostProgram exposing (Model, Msg, create)
 
 {-| The FrameRouter module is the Elm code that backs the frame-router custom element
 in the iframe-coordinator toolkit. It handles mapping URL routes to clients displayed
 in a child frame as well as message validation and routing within the parent application.
 
 This module is not currently designed for stand-alone use. You should instead use the
-custom elements defined in LINK_TO_JS_LIB to create seamless iframe applications
+custom elements defined in LINK\_TO\_JS\_LIB to create seamless iframe applications
```
justinaray: This is elm-format or vscode doing this.  Can clean up manually.

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -25,6 +42,15 @@ class FrameRouterElement extends HTMLElement {
...
+      if (labeledMsg.msgType === TOAST_REQUEST_MSG_TYPE) {
+        this.dispatchEvent(new CustomEvent(TOAST_REQUEST_EVENT_TYPE, {detail: labeledMsg.msg}));
```
justinaray: I am currently just dispatching this custom event.  Thoughts on if we should support an `onToastRequest` attribute and/or property?

justinaray: I also need to revisit the support for CustomEvent.  This constructor isn't supported in IE, but might be handled in a polyfill.

js/iframe-coordinator/libs/client.ts
```diff
@@ -23,10 +23,45 @@ function start(expectedOrigin) {
...
+     * @example
+     * worker.requestToast('Hello world');
+     * 
+     * @example
+     * worker.requestToast('World', {title: 'Hello'});
+     * 
+     * @example
+     * worker.requestToast('World', {title: 'Hello', ttl: 5, level: 'info'});
+     */
+    worker.requestToast = function (message: string, {title = null, ...custom}: undefined | ToastOptions = {}): void {
```
justinaray: Originally, the signature was `(message, title=null, custom={})`.

The more I thought about it though, the more I liked a bag of options.  Makes adding custom props easier, especially if the toast is untitled.  Open for opinions here.

The other minor issue with this is that you pass the title in as an object, but it's fired out as a first class property of the event.  `event.detail.[title|message|custom]`.  I can smooth this out either direction, but wanted to open it up to the team.

justinaray: I should probably take a look at our typescript/babel settings just to make sure we're transpiling everything I'm expecting.

js/demo-app/src/component.js
MattCheely: I think what you've got is fine. We can fancy it up if we decide to turn the app into some sort of living documentation after the POC completes. For now, I think we're better off moving forward to other features.

js/iframe-coordinator/libs/client.ts
MattCheely: Hopefully we can specify it nicely in typescript and tsdoc can make it clear for us.

js/iframe-coordinator/libs/client.ts
MattCheely: I think a signature of `(message, {title = null, custom = null})` would be better. I'm fine putting custom data in the same structure as other optional parameters, but let's give it it's own dedicated space in the object so we can evolve the API without breaking it.

I'm actually wondering if it might not be better to just expose something that doesn't translate the messages into Elm so much, but I think I will defer that question until we start to think about versioning the API.

src/HostProgram.elm
MattCheely: it's elm-format. Don't worry about it, this stuff will get changed later anyway.

PR #15, opened 2018-09-25
Routing changes.

j2pichon:


js/demo-app/public/index.html
```diff
@@ -14,11 +14,55 @@
...
+        <div class="nav">
+            <span>Programmatically:</span>
+            <br/>
+            <button class="nav-item" onclick="setRoute('#/one')">Component 1</button>
+            <button class="nav-item" onclick="setRoute('#/two')">Component 2</button>
         </div>
         <frame-router id="router"></frame-router>
     </div>
     <script type="text/javascript" src="/app.js"></script>
+    <script>
```
justinaray: This is a nit, but you could move all this js into the app.js file.  This will make development a bit easier in editors.  

You could also use the event listener on `DOMContentLoaded` to find the toggle, router, and nav items and attache the listeners.  This will also avoid polluting the global namespace.

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -1,15 +1,27 @@
 import { Host } from "../elm/Host.elm";
+import { LabeledMsg } from "../libs/types";
+
+const ROUTE_ATTR = "route";
+
+interface SubscribeHandler {
```
justinaray: nice!  :thumbsup:

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -43,14 +59,51 @@ class FrameRouterElement extends HTMLElement {
...
         this.dispatchEvent(new CustomEvent(TOAST_REQUEST_EVENT_TYPE, {detail: labeledMsg.msg}));
+      } else if (labeledMsg.msgType === NAV_REQUEST_MSG_TYPE) {
+        let event = new CustomEvent(NAV_REQUEST_EVENT_TYPE, { detail: labeledMsg.msg });
+        this.dispatchEvent(event);
       } else {
         console.error('Unexpected Message from Host Program', labeledMsg.msgType);
       }
     });
+
+    this.router.ports.toHost.subscribe(message => {
```
justinaray: Is this a dup of the above handler? ☝️

j2pichon: Indeed it is, I'll fix that :D

js/demo-app/public/index.html
j2pichon: I moved this part to index.js file.
I'm merging this, and we will enhance the demo-app next time :)

PR #16, opened 2018-09-25
Strip host route from client path before navigating client.

MattCheely:
Devs who want the previous behavior can include the same prefix
as part of the client's url hash as in the route.


README.md
```diff
@@ -38,7 +38,7 @@ host.registerElements();
  * path for that client. `url` is the page to load in the
  * iframe on that route. So in this example, the fragment
  * `#/one/my/path` would cause the `frame-router` element
- * to display the iframe at `//component/example1/#/one/my/path`
+ * to display the iframe at `//component/example1/#/my/path`
```
MattCheely: This is net effect of the changes below.

PR #17, opened 2018-10-01
Fixed build to push tag and version bump back to SCM

justinaray:
Add npm-utils to root .gitignore
Add .npmrc to js lib .gitignore
Remove npm-utils before cloning
Bump package version to test build

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 elm-stuff/
+npm-utils/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1 +1,2 @@
 elm-stuff/
+npm-utils/
```
justinaray: Alternatively, I could do this by checking out in different sub folders and modifying the scripts slightly.

js/iframe-coordinator/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "iframe-coordinator",
-  "version": "1.0.4",
+  "version": "1.0.9",
```
justinaray: Sorry for the bumps :(

PR #18, opened 2018-10-02
switch build output to umd for universal compatibility

rbrcurtis:


js/iframe-coordinator/webpack.config.js
```diff
@@ -1,11 +1,12 @@
 module.exports = {
   mode: "none",
   entry: {
+    index: "./libs/index.ts",
```
MattCheely: What does this do?

rbrcurtis: that file just bundles up both the client and server, and instead of exporting default (which would put those on window.default), it exports IframeCoordinator which then shows up as window.IframeCoodinator. 

I also think this is a bit more idiomatic because npm modules generally have an "entry point" defined in package.json. With client and host being separate there is no one entry point that could be used directly.

MattCheely: Hmmm. I don't see the file in the PR.

PR #19, opened 2018-10-03
Adding better types for the Elm modules

stormojm:
Always open for more improvements or reorg, but this gets us a bit closer to ideal.


js/iframe-coordinator/Elm.d.ts
```diff
@@ -1,7 +1,61 @@
 declare module "*/Client.elm" {
-  export var Client: any;
+  export interface LabeledMsg {
```
stormojm: Moved these interfaces into the modules since thats where they are really coming from.

PR #20, opened 2018-10-04
Upgrade to Elm 0.19

MattCheely:
This will give us better error messages, both at compile time and for
badly formatted messages. It also brings faster compile times, a smaller
bundle size, and a core URL type that's better for handling navigation
events.


js/demo-app/public/index.html
```diff
@@ -24,8 +24,8 @@
         <div class="nav">
             <span>Programmatically:</span>
             <br/>
-            <button class="nav-item" onclick="setRoute('#/one')">Component 1</button>
-            <button class="nav-item" onclick="setRoute('#/two')">Component 2</button>
+            <button class="nav-item" onclick="setRoute('/one')">Component 1</button>
+            <button class="nav-item" onclick="setRoute('/two')">Component 2</button>
```
MattCheely: This is a secondary effect of the change to navigation request payloads, which resulted in a new `RouteChange` message type in `AppToHost.elm`

js/demo-app/src/index.js
```diff
@@ -62,18 +62,18 @@ document.addEventListener("DOMContentLoaded", () => {
...
   document
     .getElementById("router")
     .addEventListener("navRequest", function(data) {
       if (window.routing) {
-        location.hash = data.detail.hash;
+        location.hash = data.detail.fragment;
         // On navRequest & routing mode, change url
       }
-      window.setRoute(data.detail.hash);
+      window.setRoute(data.detail.fragment);
```
MattCheely: The change from `hash` to `fragment` is the result of switching to the Elm 0.19 [Url package](https://package.elm-lang.org/packages/elm/url/latest/Url) for Urls.

js/iframe-coordinator/Elm.d.ts
```diff
@@ -1,69 +1,58 @@
-declare module "*/Client.elm" {
```
MattCheely: This looks like a lot of changes, but mostly it's me discovering that you don't have to explicitly export types that are used in other exports, and renaming `Worker` and `HostRouter` to `ClientProgram` and `HostProgram` to better match the original source of those types.

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -1,20 +1,8 @@
...
 import ClientFrame from "./x-ifc-frame";
-import { Location } from "../libs/types";
 
 const ROUTE_ATTR = "route";
 
-const TOAST_REQUEST_MSG_TYPE = "toastRequest";
-const TOAST_REQUEST_EVENT_TYPE = "toastRequest";
-const NAV_REQUEST_MSG_TYPE = "navRequest";
-const NAV_REQUEST_EVENT_TYPE = "navRequest";
-
```
MattCheely: I missed this in the last PR, but rather than having typescript in the loop on message labels and event types, this element just emits a custom DOM event based on the incoming type. This will make message versioning easier to handle, since the knowledge about types and labels is in fewer places.

js/iframe-coordinator/elements/frame-router.ts
```diff
@@ -42,7 +30,12 @@ class FrameRouterElement extends HTMLElement {
   }
 
   registerClients(clients: ClientRegistrations) {
-    this.router = Host.embed(this, clients);
+    let embedTarget = document.createElement("div");
+    this.appendChild(embedTarget);
+    this.router = Elm.Host.init({
+      node: embedTarget,
+      flags: clients
+    });
```
MattCheely: The way elm embeds in DOM element changed slightly in 0.19

js/iframe-coordinator/libs/client.ts
```diff
@@ -83,8 +84,13 @@ function start() {
...
     });
 
-    onLinkClick((location: Location) => {
-      sendMessage("navRequest", location);
+    window.addEventListener("click", event => {
+      let target = event.target as HTMLElement;
+      if (target.tagName.toLowerCase() === "a" && event.button == 0) {
+        event.preventDefault();
+        let a = event.target as HTMLAnchorElement;
+        sendMessage("navRequest", a.href);
```
MattCheely: We can just send along the URL now when intercepting links and parse it in Elm instead of generating a big object in typescript.

src/ClientProgram.elm
```diff
@@ -79,15 +79,20 @@ update :
...
             handleAppMessage ports.toHost message model
 
         HostMsg message ->
             handleHostMessage ports.toClient message model
 
+        BadMessage error ->
+            ( model
+            , logWarning
+                ("Could not parse incoming message: "
+                    ++ Decode.errorToString error
```
MattCheely: JSON decoding errors in Elm now produce structured data instead of just a string.  While I was dealing with that, I changed `Unknown` to `BadMessage` which I think more accurately represents what's going on in this case (we got pushed some data that we can't make sense of).

src/Message/AppToHost.elm
```diff
@@ -20,5 +20,5 @@ decodeFromApp =
         [ Decode.map Publish PubSub.publicationDecoder
         , Decode.map Subscribe PubSub.subscribeDecoder
         , Decode.map Unsubscribe PubSub.unsubscribeDecoder
-        , Decode.map NavRequest Navigation.decoder
+        , Decode.map RouteChange (Path.decoder |> expectLabel "routeChange")
```
MattCheely: It felt excessive to make a separate `Messages/RouteChange` file for this, but I could be convinced to do so pretty easily.

src/Message/Toast.elm
```diff
@@ -1,15 +1,15 @@
 module Message.Toast exposing (Toast, decoder, encode, label)
```
MattCheely: This was just updated to the naming conventions used elsewhere.

js/iframe-coordinator/Elm.d.ts
```diff
@@ -1,69 +1,58 @@
...
-    worker(): Worker;
-  }
+interface OutPort {
+  subscribe(handler: SubscribeHandler): void;
+}
 
-  export var Client: Client;
+interface ClientProgram {
+  ports: {
+    fromHost: {
```
justinaray: Not applicable for this PR, but should we eventually switch these names to be consistent with the message types?

js/iframe-coordinator/Elm.d.ts
```diff
@@ -1,69 +1,58 @@
...
+    toHost: OutPort;
+    fromClient: InPort;
+    toClient: OutPort;
+  };
 }
 
-declare module "*/Host.elm" {
-  export interface ClientRegistrations {}
+declare module "*/Client.elm" {
+  export var Elm: {
```
justinaray: do we need something more unique here?

MattCheely: In 0.19 Elm exports everything under a single `Elm` root object. So the uniqueness comes from the combined `Elm.Client`. If our Elm project exposed a `Foo` module, it'd be at `Elm.Foo`

MattCheely: It's a bit weird, because in webpack, we're including things at a file level, which makes importing `Elm.Client` from `Client.elm` a little bit redundant. But you could compile the Elm in other ways that led to multiple files & modules in a single bundle.

PR #21, opened 2018-10-05
Prepare library for publication.

MattCheely:
Publication to the elm package system gives us enforced semver, but
there are other requirements to meet as well, like having docs and
making sure the code doesn't rely on the Debug module, which can't be
used in an optimised build.

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -4,7 +4,19 @@
   "summary": "Tools for coordinating embedded apps via iframes.",
   "license": "MIT",
   "version": "1.0.0",
-  "exposed-modules": ["HostProgram", "ClientProgram", "ClientMessage"],
+  "exposed-modules": [
+    "Message.AppToClient",
+    "Message.AppToHost",
+    "Message.ClientToApp",
+    "Message.ClientToHost",
+    "Message.HostToApp",
+    "Message.HostToClient",
+    "Message.Navigation",
+    "Message.PubSub",
+    "Message.Toast",
+    "ClientProgram",
+    "HostProgram"
+  ],
   "elm-version": "0.19.0 <= v < 0.20.0",
   "dependencies": {
     "NoRedInk/elm-json-decode-pipeline": "1.0.0 <= v < 2.0.0",
@@ -15,5 +27,7 @@
     "elm/url": "1.0.0 <= v < 2.0.0",
     "elm-community/list-extra": "8.0.0 <= v < 9.0.0"
   },
-  "test-dependencies": {}
+  "test-dependencies": {
+    "elm-explorations/test": "1.1.0 <= v < 2.0.0"
+  }
 }
--- a/src/Message/Navigation.elm
+++ b/src/Message/Navigation.elm
@@ -1,4 +1,11 @@
-module Message.Navigation exposing (Navigation, decoder, encode, label, urlDecoder)
+module Message.Navigation exposing (Navigation, encode, decoder, urlDecoder, label)
+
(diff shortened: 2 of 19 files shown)


elm.json
```diff
@@ -4,7 +4,19 @@
...
+    "Message.AppToHost",
+    "Message.ClientToApp",
+    "Message.ClientToHost",
+    "Message.HostToApp",
+    "Message.HostToClient",
+    "Message.Navigation",
+    "Message.PubSub",
+    "Message.Toast",
+    "ClientProgram",
+    "HostProgram"
```
MattCheely: We need a major version bump if any of these change, since they're part of the API, so we need to expose them. That means the elm package manage will also require that we document them. That's why there are new docs below, which IMO makes the code better as well.

src/Message/Toast.elm
```diff
@@ -1,38 +1,62 @@
...
 
 
+{-| Decoder for Toast messages, which are expected to be tagged with
+`label`
+-}
 decoder : Decoder Toast
 decoder =
     (Decode.succeed Toast
-        |> optional "title" Decode.string ""
+        |> optional "title" (Decode.maybe Decode.string) Nothing
```
MattCheely: The decoder and encoder changed here to make testing the decode/encode cycle easier. The test is a preliminary step in showing how we can maintain backwards compatibility checks when evolving the API.

tests/Fixture/ToastMessage.elm
```diff
@@ -0,0 +1,13 @@
...
+import Json.Encode as Encode
+import Message.Toast exposing (Toast)
+
+
+toastFuzzer : Fuzzer Toast
+toastFuzzer =
+    Fuzz.map3 Toast
+        (Fuzz.maybe Fuzz.string)
+        Fuzz.string
+        (Fuzz.maybe (Fuzz.constant Encode.null))
```
MattCheely: Elm's testing library has property-based testing built in. This allows for the random generation of data to use in tests via "fuzzers". This describes a fuzzer that will create a `Toast` with random data in it. Tests using the fuzzer will test many random versions (100 by default) and if any of them fail, will try to reduce the data to a minimal failing test case. It's really nice for validating things like Encoders and Decoders where there are a lot of potential variations. There aren't a lot of possibilities here, but it may grow in the future, and it's easy enough to reduce the number of fuzz iterations used in a test.

tests/ToastMessage.elm
```diff
@@ -0,0 +1,18 @@
...
+
+suite : Test
+suite =
+    describe "Toast Messages"
+        [ fuzz toastFuzzer "Encodes and decodes consistently" <|
+            \toast ->
+                Toast.encode toast
+                    |> decodeValue Toast.decoder
+                    |> Expect.equal (Ok toast)
+        ]
```
MattCheely: This test generates 100 random instances of a `Toast`, round-trips them through JSON, and validates that the output matches the input, ensuring that decoding and encoding match up.

src/Message/Navigation.elm
```diff
@@ -7,15 +14,23 @@ import LabeledMessage exposing (expectLabel, withLabel)
...
 
+{-| Navigation requests are represented with an alias to the Url type.
+-}
 type alias Navigation =
     Url
 
 
+{-| This is the label that navigation events are tagged with when serialized
+to JSON. Other modules should not need to reference it, but it is exposed to
+force a package version bump if it changes.
```
justinaray: :thumbsup:

PR #22, opened 2018-10-06
Versioning test v2 - DO NOT MERGE

MattCheely:
This is an example of how Elm enforces semver, and how decoders are good for maintaining backward compatibility with previous message formats. This PR is from another repo where I've previously published the package as in the `versioning-test` branch to the Elm ecosystem under a different name. Here's what happens if I just try to publish it again:

```
▶ elm publish
  Verifying MattCheely/tryframe-coordinator 2.0.0 ...

    ● Found README.md
    ● Found LICENSE
    ✗ Version 2.0.0 is not correct!

  -- ALREADY PUBLISHED -----------------------------------------------------------

  Version 2.0.0 has already been published. You cannot publish it again!
  Try using the `bump` command:

      elm bump

  It computes the version number based on API changes, ensuring that no breaking
  changes end up in PATCH releases!
```

It warns me I have to run a command to update the version. Before that, I can run `elm diff` to see the API changes:

```
▶ elm diff
This is a MAJOR change.

---- Message.Toast - MAJOR ----

    Changed:
      - type alias Toast =
            { title : Maybe String, message : String, custom : Maybe Value }
      + type alias Toast =
            { title : String.String
            , message : String.String
            , custom : Maybe.Maybe Json.Decode.Value
            , icon : String.String
            }
```

It shows the changes, which are clearly breaking. The `title` field is no longer optional, and there's a new `icon` field which is also required. Running `elm bump` yields: 

```
▶ elm bump
Based on your new API, this should be a MAJOR change (2.0.0 => 3.0.0)
Bail out of this command and run 'elm diff' for a full explanation.

Should I perform the update (2.0.0 => 3.0.0) in elm.json? [Y/n]
Version changed to 3.0.0!
```

From here, it's a matter of committing the changes, tagging the commit, and the package can be published. 

The big picture idea here is that we can base the npm package's version off of the Elm package's version, so that we always catch breaking changes. The downside is that I think it is possible that Elm may at times consider a change which _could_ be minor from the JS side as major within the context of Elm.

This doesn't solve every problem with compatibility. We can still break APIs in a way that will cause problems for out-of-sync deploys (like changing the label), but I don't know of a set of tools for TypeScript that provides as much automatically, with both compile-time and runtime-checks.

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -1,33 +1,33 @@
 {
-  "type": "package",
-  "name": "MattCheely/tryframe-coordinator",
-  "summary": "Tools for coordinating embedded apps via iframes.",
-  "license": "MIT",
-  "version": "1.0.0",
-  "exposed-modules": [
-    "Message.AppToClient",
-    "Message.AppToHost",
-    "Message.ClientToApp",
-    "Message.ClientToHost",
-    "Message.HostToApp",
-    "Message.HostToClient",
-    "Message.Navigation",
-    "Message.PubSub",
-    "Message.Toast",
-    "ClientProgram",
-    "HostProgram"
-  ],
-  "elm-version": "0.19.0 <= v < 0.20.0",
-  "dependencies": {
-    "NoRedInk/elm-json-decode-pipeline": "1.0.0 <= v < 2.0.0",
-    "elm/browser": "1.0.0 <= v < 2.0.0",
-    "elm/core": "1.0.0 <= v < 2.0.0",
-    "elm/html": "1.0.0 <= v < 2.0.0",
-    "elm/json": "1.0.0 <= v < 2.0.0",
-    "elm/url": "1.0.0 <= v < 2.0.0",
-    "elm-community/list-extra": "8.0.0 <= v < 9.0.0"
-  },
-  "test-dependencies": {
-    "elm-explorations/test": "1.1.0 <= v < 2.0.0"
-  }
+    "type": "package",
+    "name": "MattCheely/tryframe-coordinator",
+    "summary": "Tools for coordinating embedded apps via iframes.",
+    "license": "MIT",
+    "version": "2.0.0",
+    "exposed-modules": [
+        "Message.AppToClient",
+        "Message.AppToHost",
+        "Message.ClientToApp",
+        "Message.ClientToHost",
+        "Message.HostToApp",
+        "Message.HostToClient",
+        "Message.Navigation",
+        "Message.PubSub",
+        "Message.Toast",
+        "ClientProgram",
+        "HostProgram"
+    ],
+    "elm-version": "0.19.0 <= v < 0.20.0",
+    "dependencies": {
+        "NoRedInk/elm-json-decode-pipeline": "1.0.0 <= v < 2.0.0",
+        "elm/browser": "1.0.0 <= v < 2.0.0",
+        "elm/core": "1.0.0 <= v < 2.0.0",
+        "elm/html": "1.0.0 <= v < 2.0.0",
+        "elm/json": "1.0.0 <= v < 2.0.0",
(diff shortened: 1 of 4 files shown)


elm.json
```diff
@@ -1,33 +1,33 @@
 {
-  "type": "package",
```
MattCheely: You can ignore these changes. They're a result of publishing under a different name, and whitespace changes from how the `elm` commands format the file.

src/Message/Toast.elm
```diff
@@ -21,9 +21,10 @@ app-specific features. Common use cases for app-specific features are things
 like timed auto-dismissal, notification levels or icons, etc.
 -}
 type alias Toast =
-    { title : Maybe String
+    { title : String
     , message : String
     , custom : Maybe Decode.Value
+    , icon : String
```
MattCheely: Breaking the toast API by changing the required data to construct a `Toast`

src/Message/Toast.elm
```diff
@@ -41,21 +42,42 @@ label =
 -}
 decoder : Decoder Toast
 decoder =
+    Decode.oneOf
+        [ v2decoder
+        , v1Decoder
+        ]
```
MattCheely: To handle old versions, we create decoders for previous supported versions and use `Decode.oneOf` to just try them in order.

src/Message/Toast.elm
```diff
@@ -41,21 +42,42 @@ label =
...
 
+v1Decoder : Decoder Toast
+v1Decoder =
+    (Decode.succeed Toast
+        |> optional "title" Decode.string "Notification!"
+        |> required "message" Decode.string
+        |> optional "custom" (Decode.maybe Decode.value) Nothing
+        |> hardcoded "info"
+    )
+        |> expectLabel "toastRequest"
```
MattCheely: The v1 decoder provides default & fixed values for fields that were previously optional or not present.

src/Message/Toast.elm
```diff
@@ -41,21 +42,42 @@ label =
...
+
 {-| Encodes a Toast to JSON, tagging it with `label`
 -}
 encode : Toast -> Encode.Value
 encode toast =
-    [ Maybe.map (\title -> ( "title", Encode.string title )) toast.title
+    [ Just ( "title", Encode.string toast.title )
     , Just ( "message", Encode.string toast.message )
     , Maybe.map (\custom -> ( "custom", custom )) toast.custom
+    , Just ( "icon", Encode.string toast.icon )
```
MattCheely: `encode` only needs to support the new format.

tests/Fixture/ToastMessage.elm
```diff
@@ -1,13 +1,51 @@
...
+        (Fuzz.oneOf
+            [ Fuzz.constant "info"
+            , Fuzz.constant "warning"
+            , Fuzz.constant "error"
+            ]
+        )
+
+
+
+-- Old Versions
```
MattCheely: To test support for old messages, I just copied over the old `Toast` type and renamed it to `ToastV1`, and also moved the old fuzzer and encoder to generate JSON in the old format.

tests/ToastMessage.elm
```diff
@@ -15,4 +15,17 @@ suite =
...
+            \v1Toast ->
+                Fixture.encodeV1 v1Toast
+                    |> decodeValue Toast.decoder
+                    |> Expect.equal
+                        (Ok
+                            (Toast
+                                (v1Toast.title |> Maybe.withDefault "Notification!")
+                                v1Toast.message
+                                v1Toast.custom
+                                "info"
```
MattCheely: To validate old versions, we generate a bunch randomly, decode them, and verify that the input data shows up correctly in the generated `Toast` object, or that the correct defaults are used.

PR #24, opened 2018-10-11
Converting client into a class structure.

stormojm:
Pretty simple rework, just to make sure the idea/coding style is agreed upon.  I also want to add some tests as part of this, since I wish I had that to confirm everything worked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,12 +25,12 @@ import "@webcomponents/custom-elements/src/native-shim.js";
 import "@webcomponents/custom-elements/src/custom-elements.js";
 
 /* Import the host library */
-import host from "iframe-coordinator/host.js";
+import { registerElements } from "iframe-coordinator/host.js";
 
 /* This will register two custom elements, `frame-router` and
  * `component-frame`. Only frame-router should be used directly.
  */
-host.registerElements();
+registerElements();
 
 /* This registers three client apps with the `frame-router`
  * element with an id of `coordinator`. `assignedRoute` is the
@@ -70,7 +70,10 @@ document.getElementById("frame-element").registerClients({
 
 ```js
 /* Import the client library */
-import client from "iframe-coordinator/client.js";
+import { Client } from "iframe-coordinator/client.js";
+
+/* Create a new instance of the client */
+const client = new Client();
 
 /* Start intercepting links, which will prevent the
  * default link action, and instead send a message to the
--- a/js/iframe-coordinator/libs/client.ts
+++ b/js/iframe-coordinator/libs/client.ts
@@ -3,70 +3,110 @@ import { PublicationHandler } from "./types";
 
 interface ToastOptions {
   title?: string;
-  custom?: { [x: string]: any };
+  custom?: any;
+}
+
+interface ClientConfigOptions {
+  clientWindow?: Window
 }
 
 class Client {
   private _worker: ClientProgram;
+  private _isStarted: boolean;
+  private _clientWindow: Window;
   private _messageHandlers: Array<PublicationHandler> = [];
 
-  start(): void {
-    if (!this._worker) {
-      this._worker = Elm.Client.init();
-  
-      window.addEventListener("message", (event: MessageEvent) => {
-        this._worker.ports.fromHost.send({
-          origin: event.origin,
-          data: event.data
-        });
-      });
-  
-      this._worker.ports.toHost.subscribe((message: any) => {
-        window.parent.postMessage(message, "*");
-      });
-  
-      this._worker.ports.toClient.subscribe((message: any) => {
-        if (message.msgType == "publish") {
-          this._messageHandlers.forEach(handler => {
-            handler(message.msg);
-          });
-        }
-      });
-  
-      window.addEventListener("click", event => {
-        let target = event.target as HTMLElement;
-        if (target.tagName.toLowerCase() === "a" && event.button == 0) {
-          event.preventDefault();
-          let a = event.target as HTMLAnchorElement;
-          this._sendMessage("navRequest", a.href);
-        }
-      });
+  public constructor(configOptions: ClientConfigOptions = {}) {
+    this._clientWindow = (configOptions.clientWindow) || window;
+    this._worker = Elm.Client.init();
+  }
+
+
+  private _sendingMessageToHost = (message: LabeledMsg) => {
+    this._clientWindow.parent.postMessage(message, "*");
+  }
+
+  private _publishMessageToHandlers = (message: LabeledMsg) => {
+    if (message.msgType != "publish") {
+      return;
+    }
+
+    this._messageHandlers.forEach(handler => {
+      handler(message.msg);
+    });
+  }
+
+  private _onWindowMessageReceived = (event: MessageEvent) => {
+    this._worker.ports.fromHost.send({
+      origin: event.origin,
+      data: event.data
+    });
+  }
+
+  private _onWindowClick = (event: MouseEvent) => {
+    let target = event.target as HTMLElement;
+    if (target.tagName.toLowerCase() === "a" && event.button == 0) {
+      event.preventDefault();
+      let a = event.target as HTMLAnchorElement;
+      this._sendMessage("navRequest", a.href);
     }
   }
 
-  _sendMessage(type: string, data: any): void {
+  public start(): void {
+    if (this._isStarted) {
+      return;
+    }
+
+    this._isStarted = true;
+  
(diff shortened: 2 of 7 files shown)


js/iframe-coordinator/libs/client.ts
```diff
@@ -1,31 +1,74 @@
...
+        window.parent.postMessage(message, "*");
+      });
+  
+      this._worker.ports.toClient.subscribe((message: any) => {
+        if (message.msgType == "publish") {
+          this._messageHandlers.forEach(handler => {
+            handler(message.msg);
+          });
+        }
+      });
```
MattCheely: Do you think it would improve the API to handle the call to `Elm.Client.init()` in the constructor, along with the port setup, and then maybe call this something like `startLinkHandler`?  I don't think there's a real benefit to deferring the Elm setup steps, but I could see a case where a client app might not want to activate the link handling by default.

MattCheely:
Regarding coding style, I'm a big fan of auto-formatters. What would y'all say to setting up a [prettier configuration](https://prettier.io/docs/en/configuration.html) for the project?

js/iframe-coordinator/libs/client.ts
stormojm: I think it makes sense to leave any eventing behind some sort of init/start method.  People may want to defer the iframe from doing anything for a certain trigger.

My ideal API, i think would look something like this:
import { Client } from 'iframe-coordinator';

if (enabled) {
  client.start({ linkHandlers: true });
}

onDisabled() {
  client.stop();
}


either that or Client doesn't return an object, but a subscription func:
import { client } from 'iframe-coordinator';

if (enabled)
  client({ link:handlers: true });

stormojm:
I like auto-formatters as well.  Ideally we use something consistent between all of our projects.  I would also hope we would use something like tslint as well.

js/iframe-coordinator/libs/client.ts
```diff
@@ -53,56 +96,13 @@ export default {
...
-    msgType: type,
-    msg: data
-  });
 }
 
-interface ToastOptions {
-  title?: string;
-  custom?: { [x: string]: any };
-}
+const client = new Client();
```
justinaray: Maybe we should make this `Client.ts` and export the class itself.  This is a pretty common approach and then we have the ability to instantiate multiple instances for testing, etc.

js/iframe-coordinator/libs/client.ts
justinaray: Mentioned something similar below, but my vote is:

1. Export the class
1. high-level config can be passed into the constructor
1. init/start function to delay execution as desired (fine if we put listening type config here and not in the constructor)

js/iframe-coordinator/libs/client.ts
```diff
@@ -1,31 +1,74 @@
...
+          this._sendMessage("navRequest", a.href);
+        }
+      });
+    }
+  }
 
-//TODO: make this a class, for a more idiomatic API?
-export default {
-  start: start,
+  _sendMessage(type: string, data: any): void {
```
justinaray: We should make this `private` in addition to the _ convention.

js/iframe-coordinator/libs/client.ts
MattCheely: I'm fine with keeping all of the initialization in the `start` method if ya'll think it's better that way. My initial comment was made assuming that `client.ts` exported the Client constructor, not an instance. I missed that detail the first time around. I do think exporting the class rather than just an instance is a good idea. That lets consumers extend it if they want. If we want to enforce a singleton client instance, we could do that within the constructor.

MattCheely:
I really wish github had a better way of threading top-level conversations. @stormojm - I mentioned it this morning, but yeah, we should get a prettier config and tslint setup that's shared across the two common UI projects. We could give those their own package to de-dupe, but copy-and-pasting across repos is probably also not too bad, at least for now.

js/iframe-coordinator/libs/client.ts
stormojm: sounds good to me.  I didn't know if there was a reason we made it a singleton in the first place.

README.md
```diff
@@ -25,12 +25,12 @@ import "@webcomponents/custom-elements/src/native-shim.js";
 import "@webcomponents/custom-elements/src/custom-elements.js";
 
 /* Import the host library */
-import host from "iframe-coordinator/host.js";
+import { registerElements } from "iframe-coordinator/host.js";
```
stormojm: This felt so much better to me in terms of a cleaner API.  It felt really odd to get an object just to call this one method.

js/iframe-coordinator/libs/client.ts
```diff
@@ -3,70 +3,110 @@ import { PublicationHandler } from "./types";
...
-  
-      this._worker.ports.toClient.subscribe((message: any) => {
-        if (message.msgType == "publish") {
-          this._messageHandlers.forEach(handler => {
-            handler(message.msg);
-          });
-        }
-      });
+  public constructor(configOptions: ClientConfigOptions = {}) {
+    this._clientWindow = (configOptions.clientWindow) || window;
```
stormojm: As discussed in the unit testing pull request.

js/iframe-coordinator/libs/client.ts
```diff
@@ -3,70 +3,110 @@ import { PublicationHandler } from "./types";
...
+    this._worker = Elm.Client.init();
+  }
+
+
+  private _sendingMessageToHost = (message: LabeledMsg) => {
```
stormojm: This is annoying part about js/ts classes.  Methods don't bind this automatically.  So instead we use private fields which do with lambdas, so that we can add/remove the same instance.

I think this is the best of the available options, but I am also ok with making these methods again and having fields for the bound versions of the methods.  (Or suggest something else).

js/iframe-coordinator/libs/client.ts
```diff
@@ -3,70 +3,110 @@ import { PublicationHandler } from "./types";
...
-          this._sendMessage("navRequest", a.href);
-        }
-      });
+    this._clientWindow.addEventListener("message", this._onWindowMessageReceived);
+    this._clientWindow.addEventListener("click", this._onWindowClick);
+    this._worker.ports.toHost.subscribe(this._sendingMessageToHost);
+    this._worker.ports.toClient.subscribe(this._publishMessageToHandlers);
+  }
+
+  public stop(): void {
```
stormojm: Added stop for event cleanup and just a more complete API.

js/iframe-coordinator/libs/client.ts
```diff
@@ -104,5 +144,4 @@ class Client {
   }
 }
 
-const client = new Client();
-export default client;
\ No newline at end of file
+export { Client };
```
stormojm: Exporting the constructor now.

js/iframe-coordinator/libs/index.ts
```diff
@@ -1,5 +1,4 @@
-import client from './client';
```
stormojm: I don't think this is even used in our module now.  I personally still think it would be better to just export index and let them consume It like `import {Client} from 'iframe-coordinator`

js/iframe-coordinator/libs/client.ts
stormojm: Btw, I really am disliking the name "Client".  I purpose we may want to name it something else.  Maybe something similar to Parcel or Fragment like mosaic/single-spa uses for their micro-frontends?

MattCheely: I used to have a different set of names for the parent app and the embedded app, something like "coordinator" and "component" but after taking an informal survey (me walking around the office asking folks), it seemed like the terms "host" and "client" for the parent app and the iframe were the things that people found easiest to understand quickly which thing was which in the system.  Do you think Client is too generic? It's certainly an overloaded term and I could see how it could be confusing in the context of an app that might have clients for other things.

stormojm: Yea, I think its too generic and overloaded.  I just feel we could give it something more descriptive to its purpose.

js/iframe-coordinator/libs/index.ts
MattCheely: I agree that that's a nicer API for importing. My only question is would that result in the host & client code getting bundled with every application, or the latest approaches to tree-shaking able to recognize that the host code in `index` is never used and exclude it from the build? I didn't think they were capable of that last time I looked, but I haven't done a deep dive into it.

PR #25, opened 2018-10-11
Document some modal API ideas. 

MattCheely:
Let's try a PR for async discussion across time zones.

### [Rendered Document](https://github.com/purecloudlabs/iframe-coordinator/blob/modal-api/modal-api.md)

### DIFF
--- a/modal-api.md
+++ b/modal-api.md
@@ -1,0 +1,58 @@
+# Design ideas for modal dialogs
+
+## The simple approach
+
+Several of the UX modal designs are pretty simple, consisting of a title, some text and some buttons (examples). These prompt-like intefaces could be supported with a cross-frame message with the format:
+
+```TypeScript
+interface Prompt = {
+  title : string,
+  message : string,
+  button : array<Button>
+}
+
+interface Button = {
+  text : string,
+  type : 'secondary' | 'primary',
+  id : string
+}
+```
+
+When the host app displays the prompt and the user clicks a button, the client would be sent a message:
+
+```TypeScript
+interface ModalResult = {
+  buttonId : string
+}
+```
+
+With the caveat that buttonId can probably always be 'cancel', if the user closes the modal.
+
+## The complicated approach
+
+The more complex option is that when it wants to display a modal, the client app sends a URL to the parent app via a message like:
+
+```TypeScript
+interface Modal = {
+  frame: string
+}
+```
+
+example:
+
+```TypeScript
+{
+  frame: "https://www.com/theBestApp/groupEditModal.html?groupId=1234"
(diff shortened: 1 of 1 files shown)


modal-api.md
```diff
@@ -0,0 +1,58 @@
...
+
+## The simple approach
+
+Several of the UX modal designs are pretty simple, consisting of a title, some text and some buttons (examples). These prompt-like intefaces could be supported with a cross-frame message with the format:
+
+```TypeScript
+interface Prompt = {
+  title : string,
+  message : string,
+  button : array<Button>
```
stormojm: We may have to account for the UX guidelines seperation of the left-side and right-side buttons.  Unless we assume the first or last is the right side.

modal-api.md
```diff
@@ -0,0 +1,58 @@
+# Design ideas for modal dialogs
+
+## The simple approach
+
+Several of the UX modal designs are pretty simple, consisting of a title, some text and some buttons (examples). These prompt-like intefaces could be supported with a cross-frame message with the format:
+
+```TypeScript
+interface Prompt = {
+  title : string,
```
justinaray: This is getting into implementation details, but will want the modal requestor to have a few more options for behavior:

closable, clickOutsideToClose.

The host app can use these properties and it's own styles to decide if the model has a glasspane behind it and we can handle iframe blur events to support click outside.

modal-api.md
```diff
@@ -0,0 +1,58 @@
...
+
+When the host app displays the prompt and the user clicks a button, the client would be sent a message:
+
+```TypeScript
+interface ModalResult = {
+  buttonId : string
+}
+```
+
+With the caveat that buttonId can probably always be 'cancel', if the user closes the modal.
```
justinaray: Per my comment above, we'll might want to explicitly track the close method.  button, clickoutside, and close.  Generally, these are handled the same, but I've definitely had projects in the past where I wanted to handle these cases differently.

modal-api.md
```diff
@@ -0,0 +1,58 @@
...
+example:
+
+```TypeScript
+{
+  frame: "https://www.com/theBestApp/groupEditModal.html?groupId=1234"
+}
+```
+
+The host app will open a modal dialog and display the url in an iframe as part of the modal. The iframe can do
+whatever it needs and when it's complete, send a message to the host app, which will be passed back to the client:
```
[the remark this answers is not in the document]
justinaray: Yeah, it would be cool to mimic the planned init api here so that we could seed some data to the modal.  Then it would be app-specific if they wanted to use cached data or fetch the most up-to-date info.

PR #26, opened 2018-10-11
adding same styling for programmatic routing

j2pichon:
buttons will look like tabs ;)

### DIFF
--- a/js/demo-app/src/index.js
+++ b/js/demo-app/src/index.js
@@ -150,17 +150,17 @@ function toggleRouting() {
 
   if (urlRoutingEnabled) {
     setRoute(window.location.hash.slice(1));
+    updateActiveNav();
   } else {
     location.hash = "";
     setRoute('');
-    updateActiveNav();
+    updateActiveNav('');
   }
 
   // Update UI
   document.querySelector('button.url-routing.toggle-switch').setAttribute('aria-checked', urlRoutingEnabled);
   document.querySelector('header nav ul.nav-menu').style.display = urlRoutingEnabled ? 'flex' : 'none';
   document.querySelector('header nav ul.programmatic-nav').style.display = urlRoutingEnabled ? 'none' : 'flex';
-  updateActiveNav();
 };
 
 // UI Helpers
--- a/js/demo-app/src/index.js
+++ b/js/demo-app/src/index.js
@@ -92,6 +92,10 @@ document.addEventListener("DOMContentLoaded", () => {
         // On navRequest & routing mode, change url
       } else {
         setRoute(data.detail.fragment);
+        let activeNavItem = NAV_CONFIGS.find(toMatch => {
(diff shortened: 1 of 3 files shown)


js/demo-app/src/index.js
```diff
@@ -183,7 +187,11 @@ function buildNavMarkup(navConfigs) {
     // Build Programmatic Button
     let currButton = document.createElement('button');
     currButton.setAttribute('class', `nav-id-${curr.id}`);
-    currButton.addEventListener('click', () => {
+    currButton.addEventListener('click', (e) => {
+      programmaticNav.querySelectorAll('button').forEach(el => {
```
justinaray: Should we move this code into `updateActiveNav` so it's in one place.  Then it can be called here on click and on routing events from the frame coordinator?

I left the fqRoute (poorly named) as a param to that method so that we could eventually support other types of active tracking

js/demo-app/src/index.js
```diff
@@ -181,13 +187,17 @@ function buildNavMarkup(navConfigs) {
     navMenu.appendChild(currLi);
 
     // Build Programmatic Button
+    let currLinkButton = document.createElement('li');
+    currLinkButton.setAttribute('class', `nav-id-${curr.id}`);
     let currButton = document.createElement('button');
     currButton.setAttribute('class', `nav-id-${curr.id}`);
     currButton.addEventListener('click', () => {
+      updateActiveNav(NAV_CONFIGS, curr.id);
```
justinaray: D'oh, too much functional!  I guess i should have made the navs the second argument so we only have have to pass the custom route in this case and no params in the other.  Feel free to switch this around if you want.

js/demo-app/src/index.js
```diff
@@ -92,6 +92,10 @@ document.addEventListener("DOMContentLoaded", () => {
         // On navRequest & routing mode, change url
       } else {
         setRoute(data.detail.fragment);
+        let activeNavItem = NAV_CONFIGS.find(toMatch => {
```
justinaray: I think it would work if you pass `data.detail.fragment` into updateActiveNav and just remove the check on `urlRoutingEnabled` in that method.

I haven't tested it, but that was my intention when I added that string param.  Sorry that i named it `fqRouteName`.  That wasn't the best

js/demo-app/src/index.js
```diff
@@ -201,18 +211,21 @@ function updateActiveNav(navConfigs = NAV_CONFIGS, fqRoute = window.location.has
       let activeNavItem = navConfigs.find(toMatch => {
         return (toMatch.assignedRoute === currRoute);
       });
-
       activeNavId = (activeNavItem ? activeNavItem.id : null);
     }
+  } else {
```
justinaray: Here's where I was talking.  If you pull out this conditionally and just search the fqRoute, you should be able to use it for programmic or url versions.

PR #27, opened 2018-10-12
Adding initial unit testing infrastructure

stormojm:


js/iframe-coordinator/karma.conf.js
```diff
@@ -0,0 +1,26 @@
+let webpackConfig = require('./webpack.config');
+
+module.exports = function(config) {
+  config.set({
+    basePath: '',
+    frameworks: ['jasmine'],
```
stormojm: Are we ok with jasmine?

js/iframe-coordinator/karma.conf.js
```diff
@@ -0,0 +1,26 @@
...
+    preprocessors: {
+      '**/*.spec.ts': ['webpack']
+    },
+    webpack: webpackConfig,
+    reporters: ['progress'],
+    port: 9876,
+    colors: true,
+    logLevel: config.LOG_INFO,
+    autoWatch: true,
+    browsers: ['Chrome'],
```
stormojm: I am not sure how testing is normally done with our builds.  If we can use chrome, or if we have to use headless.  Let me know whats the best practice.

js/iframe-coordinator/libs/client.ts
```diff
@@ -61,19 +61,23 @@ export default {
   }
 };
 
-function start() {
+function start(frameWindow: Window) {
```
stormojm: This is probably the most controversial thing.

I wasn't allowed to set global.window (its readonly). So if we want to mock window, we have to inject the dependencies.  Now that I think about it more in the review, I think we may want to make this a config object instead of a parameter.  But for now i'll leave this up for comments/alternatives.

js/iframe-coordinator/libs/specs/client.spec.ts
```diff
@@ -0,0 +1,245 @@
+import * as ClientInjector from 'inject-loader!../client';
+
+describe("client", () => {
+  let client: any,
```
stormojm: I want to explore mocking in another pull-request since I don't like having these be any-typed.

js/iframe-coordinator/package.json
```diff
@@ -12,14 +13,22 @@
...
     "source-map-loader": "^0.2.4",
     "ts-loader": "^5.0.0",
     "typescript": "^3.0.3",
     "webpack": "^4.16.3",
     "webpack-cli": "^3.1.0"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "./node_modules/karma/bin/karma start --single-run",
+    "start_test": "./node_modules/karma/bin/karma start",
```
stormojm: With this config, we aren't able to do the karma cli "karma start:".  It fails trying to spawn elm, for some reason.  So i  just made this npm helper (i hate the name, so give me some better ones :))

js/iframe-coordinator/karma.conf.js
justinaray: We usually install chromium-headless for CI runs.

In some of our projects, we use qunit which differentiates between desktop and ci runs.  It uses Chromium/Chrome local install in the former and then chromium headless in the ci runs.

Although the index.js file is specifically for Ember/QUnit, you may be able to just add this library as a dependency and use some of the scripts to check/install.

https://github.com/purecloudlabs/ember-chromium

js/iframe-coordinator/libs/client.ts
justinaray: Yeah, no getting around the need to mock window.  Been there, go the t-shirt ;)

I've definitely done it like this and passed it in with `window` being the default value.

I've also used a `_window` "private" variable before.  I don't make it a  proper `private` but rely on the underscore convention for people not to use it.  Then, I can set the property to a mocked object in my tests and not pollute the API contract with testing concerns.

Both will work fine though.

js/iframe-coordinator/karma.conf.js
MattCheely: I've never found a JS unit testing library that I particularly liked, so... yes? :stuck_out_tongue_winking_eye:

js/iframe-coordinator/karma.conf.js
MattCheely: We'll definitely need a headless option for CI builds.

js/iframe-coordinator/libs/client.ts
MattCheely: I'm fine with any approach to enable mocking window. My only strong opinion here is that under normal usage, consumers shouldn't have to explicitly pass in the window object.

js/iframe-coordinator/package.json
MattCheely: I will pull this down locally and see if I can figure out what's going on with the Elm integration.

MattCheely: Were you running a globally installed karma for `karma start`? I tried it and got different failure:

```
12 10 2018 14:12:28.473:ERROR [preprocess]: Can not load "webpack", it is not registered!
  Perhaps you are missing some plugin?
12 10 2018 14:12:28.477:ERROR [karma]: Error: No provider for "framework:jasmine"! (Resolving: framework:jasmine)
```
Oddly, running `./node_modules/.bin/karma start` works fine. I'm guessing that when it's running from the global install it's looking for other globally installed tools, but when running from `node_modules` it looks in the local `node_modules`.  I don't have webpack installed globally, and you probably don't have the elm compiler installed globally. 

Personally, I like running from the npm scripts better. It makes sure everybody is using the same version of karma to run tests. 

`start_test` looks like it re-runs tests automatically? In which case maybe call it `test-watch` or `autotest`?  (I have a mild preference for hyphens over underscores, but it's mild, so ¯\_(ツ)_/¯)

js/iframe-coordinator/karma.conf.js
stormojm: Alright, I'll set get that up here on Monday.

js/iframe-coordinator/libs/client.ts
stormojm: I'll leave this here for now, until i merge in the class changes from the other pr.  That will move this into the constructor and change it up a bit anyways.

js/iframe-coordinator/package.json
```diff
@@ -13,13 +13,15 @@
...
     "elm-webpack-loader": "^5.0.0",
+    "i": "^0.3.6",
     "import-inject-loader": "^0.2.4",
     "inject-loader": "^4.0.1",
     "jasmine-core": "^3.2.1",
     "karma": "^3.0.0",
     "karma-chrome-launcher": "^2.2.0",
     "karma-jasmine": "^1.1.2",
     "karma-webpack": "4.0.0-rc.2",
+    "puppeteer": "^1.9.0",
```
stormojm: Using puppeteer for browser (I assumed the build machines wouldn't have chrome installed)

PR #29, opened 2018-10-18
Generating the definition files for consumption by users of this libr…

stormojm:
…ary.  It annoying that we have to pollute the root space, but while we have multiple entries, thats the only option

You can read more info 
https://github.com/Microsoft/TypeScript/issues/8305
https://www.chriskrycho.com/2018/how-to-bundle-typescript-type-definitions.html


js/iframe-coordinator/tsconfig.json
```diff
@@ -3,9 +3,10 @@
...
     "module": "commonjs",
-    "target": "es5",
-    "allowJs": true
+    "declaration": true,
+    "declarationDir": "./",
+    "target": "es5"
   },
-  "include": ["**/*.ts"],
-  "exclude": ["node_modules"]
+  "include": ["libs"],
```
stormojm: This helps typescript to generate the files into the root dir.  If you use the glob pattern, it keeps the src directory.

The alternative would to have a copy step to move them to root.

js/iframe-coordinator/webpack.config.js
```diff
@@ -1,3 +1,5 @@
+const path = require('path');
```
stormojm: I am going to revert this.

js/iframe-coordinator/libs/index.ts
```diff
@@ -1,4 +1,4 @@
 import { Client } from './client';
-import { registerElements } from './host';
+import { registerElements, FrameRouterElement } from './host';
```
stormojm: Exported this as well, to allow consumers to reference the type.

PR #30, opened 2018-10-19
Adding prettier and tslint configurations (and updating files to match)

stormojm:
Adding tslint and prettier configurations which were originally pulled from the web components repository.


js/iframe-coordinator/tslint.json
```diff
@@ -0,0 +1,15 @@
+{
+    "extends": ["tslint:recommended", "tslint-config-prettier"],
+    "linterOptions": {
+      "exclude": ["node_modules/**/*.ts"]
+    },
+    "rules": {
```
stormojm: There are a few differences to the rules that I wanted to point out:
- member-access dropped no-public.  This was because Justin had originally requested u add public.  I am fine with either if we want to match.
- member ordering was added.  I like grouping related methods rather than forcing up to have public than private.
- no-trailing-whitespace was added.
- object-literal-sort-keys was added.  Having to alphabetically order keys in objects is annoying to me.
- variable-name was added.  We were using _ for private members

js/iframe-coordinator/package.json
```diff
@@ -22,19 +22,26 @@
...
     "typescript": "^3.0.3",
     "webpack": "^4.16.3",
     "webpack-cli": "^3.1.0"
   },
   "scripts": {
     "test": "./node_modules/karma/bin/karma start --single-run",
-    "test-watch": "./node_modules/karma/bin/karma start",
-    "test-watch-chrome": "./node_modules/karma/bin/karma start --browsers=Chrome",
+    "test.watch": "./node_modules/karma/bin/karma start",
+    "test.watch.chrome": "./node_modules/karma/bin/karma start --browsers=Chrome",
```
stormojm: updated this format (using . for 'options') to match the commands used in web components.

js/iframe-coordinator/tslint.json
MattCheely: I would like object-literal-sort-keys if there was a way to autoformat it.  I'd have opted for enabling member ordering, particularly since it has a fixer which means it can probably be automated via editor config. That said, it's not a big deal and there are better ways for us to spend time than bike-shedding linter options. :grin:

PR #31, opened 2018-10-19
Feature/improve demo ie support {READY_FOR_REVIEW}

justinaray:
Was able to get the demo app and library running in Edge.  IE11 app is up and limping along, but the cross-frame events are getting lost.  This was good progress so I wanted to go ahead and open this one.

### DIFF
--- a/js/demo-app/package-lock.json
+++ b/js/demo-app/package-lock.json
@@ -4,6 +4,758 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://purecloud.artifactoryonline.com/purecloud/api/npm/inin-internal-npm/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha1-BuKrGb21NThVWaq7W6WXKUgoAPg=",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "7.0.0"
+      }
+    },
+    "@babel/core": {
+      "version": "7.1.2",
+      "resolved": "https://purecloud.artifactoryonline.com/purecloud/api/npm/inin-internal-npm/@babel/core/-/core-7.1.2.tgz",
+      "integrity": "sha1-+NKpzraDKIcymntg+dA1eRQAuk4=",
+      "dev": true,
+      "requires": {
+        "@babel/code-frame": "7.0.0",
+        "@babel/generator": "7.1.3",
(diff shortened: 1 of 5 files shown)


js/demo-app/package.json
```diff
@@ -16,7 +16,18 @@
   },
   "author": "",
   "license": "MIT",
+  "browserslist": [
```
justinaray: This may not be our exact supported list, but it's a decent start.  IE10 might be opening pandora's box, so I left it off this list as a first pass.  Still need to get 11 working correctly anyway ;)

js/demo-app/src/component.js
```diff
@@ -1,3 +1,6 @@
+import "@babel/polyfill";
+import "url-polyfill";
+import "nodelist-foreach-polyfill";
```
justinaray: url-polyfill was needed for both the demo and our lib while the foreach poly was just for the demo.  I think we should ultimately come up with a list of polyfills needed for the library and document them.

Perhaps, we should even offer a few different targets.  es6 modules, umd module, and umd/min modules with polyfills.  I have done this previously with rollup and it worked out pretty well.  Will just take adding a few different webpack module exports and some main entries in package.json.  But that's for another day.

js/demo-app/package-lock.json
```diff
@@ -2626,10 +3493,15 @@
     "iframe-coordinator": {
       "version": "file:../iframe-coordinator",
       "requires": {
+        "@types/jasmine": "2.8.9",
         "elm": "0.19.0-bugfix2",
         "request": "2.88.0"
       },
       "dependencies": {
+        "@types/jasmine": {
```
justinaray: @stormojm ... I went ahead and commited this package-lock.json change that came in from the lib dependency updates.  However, I noticed that these types are listed in dependency.  Could they be moved to dev?

js/demo-app/package.json
```diff
@@ -25,9 +36,14 @@
     "webpack-dev-server": "^3.1.6"
   },
   "dependencies": {
+    "@babel/polyfill": "^7.0.0",
     "@webcomponents/custom-elements": "^1.0.9",
     "copy-webpack-plugin": "^4.5.2",
+    "core-js": "^2.5.7",
+    "custom-event-polyfill": "^1.0.6",
```
justinaray: Hat tip to @j2pichon for finding the issue on ie11.  It was missing support for the CustomEvent constructor.  This polyfill takes care of that.

Working in Edge and IE11.  woot!

PR #33, opened 2018-10-23
WIP Typescript Migration

MattCheely:
This is pretty sparse, but I think it's a decent strategy for structuring the messages that will let us use TypeScript's discriminated unions for exhaustiveness checks when we start handling messages.

@stormojm I was getting a boatload of foo.d.ts files generated everywhere when I ran webpack with --watch and also seeing changes not getting picked up. 

Also, I'm wondering if pre-commit is a little bit agressive for the tslint as opposed to pre-push.

### DIFF
--- a/js/demo-app/src/component.js
+++ b/js/demo-app/src/component.js
@@ -25,7 +25,10 @@ document.getElementById("do-publish").addEventListener("click", () => {
   let data = JSON.parse(document.getElementById("pub-data").value);
 
   //Publish the data
-  iframeClient.publish(topic, data);
+  iframeClient.publish({
+    topic: topic,
+    payload: data
+  });
 });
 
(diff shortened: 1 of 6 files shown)


MattCheely:
Y'all feel free to run with this while I'm out the rest of this week.

MattCheely:
The code will definitely need to be restructured. I was going to hold off
on that, migrating individual message paths from Elm to typescript, until
the move was complete, and then re-organize. That way I wouldn't have to
break everything and rebuild again from scratch.

The bigger issue I was seeing than having a d.ts file next to every .ts
file was that after fixing a compile-time issue in typescript, the compiler
error wouldn't be resolve when webpack re-ran due to the edit, only if I
deleted the corresponding d.ts file or completely restarted webpack.

PR #34, opened 2018-10-29
Discriminated union examples

MattCheely:
Hopefully this illustrates what I had in mind for validation and skipping the client program. I used strict null checks for exhaustiveness checking on the union, but we could also potentially use the `never` type:
https://www.typescriptlang.org/docs/handbook/advanced-types.html#exhaustiveness-checking

Either way, I think we should leave strict null checks on. We'll have fewer bugs without it.

### DIFF
--- a/js/iframe-coordinator/src/messages/ClientToHost.ts
+++ b/js/iframe-coordinator/src/messages/ClientToHost.ts
@@ -1,4 +1,20 @@
 import { LabeledPublication } from './Publication';
 import { LabeledToast } from './Toast';
+import { LabeledNavRequest } from './NavRequest';
 
-export type ClientToHost = LabeledPublication | LabeledToast;
+export type ClientToHost =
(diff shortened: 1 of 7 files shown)


js/iframe-coordinator/src/messages/ClientToHost.ts
```diff
@@ -1,4 +1,20 @@
...
+export function validate(msg: ClientToHost) {
+  //TODO: actually validate cases
+  switch (msg.msgType) {
+    case 'navRequest':
+      return msg;
+    case 'toastRequest':
+      return msg;
+    case 'publish':
+      return msg;
+  }
```
MattCheely: Handling the incoming messages should also use a switch pattern like this, IMO, which will ensure at compile time that the host and client libraries have paths for every type of inbound message and that we haven't typo'd a label anywhere.

PR #36, opened 2018-10-31
Clientprogram rewrite

stormojm:
It seems like this is ok to move up a level to the original ts-migration branch.  Unless there is a reason to keep that branch seperated?


stormojm:
I didn't realize this was already reparented

PR #37, opened 2018-10-31
Decoders

stormojm:
One big annoyance here, is I didn't see a way for us to use the type definitions with the hardcoded msgTypes (with this decoder framework).  The guard object always returned a string even when using constant.

But this works as far as i can tell.  I will need to do the same thing with ClientToHost, which is more complicated, but I wanted to get this up for review.  Feel free to merge it as well, if you get to this after I leave.


js/iframe-coordinator/src/messages/specs/ClientToHost.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+        msgType: 'publish',
+        msg: {
+          topic: 'test.topic'
+        }
+      };
+      const expectedMessage = {
+        msgType: 'publish',
+        msg: {
+          topic: 'test.topic',
+          payload: undefined
```
stormojm: This is the other annoying thing.  The decoder adds a key for optional properties even if they weren't in the original json.  They are still undefined, so its not a huge deal, but its still not expected

js/iframe-coordinator/src/messages/NavRequest.ts
```diff
@@ -5,6 +5,6 @@ export interface NavRequest {
 }
 
 export interface LabeledNavRequest extends LabeledMsg {
-  msgType: 'navRequest';
+  msgType: string;
```
stormojm: If this is ok with you, we can change the types around a bit to only have the | on the msg itself.  So ClientToHost can be msg: NavRequest | Toast | Publication

js/iframe-coordinator/src/messages/ClientToHost.ts
```diff
@@ -7,14 +16,42 @@ export type ClientToHost =
...
+  })
+});
+
+const decoder = guard(
+  either3(
+    labeledPublicationDecoder,
+    labeledToastDecoder,
+    labeledNavRequestDecoder
+  )
+);
```
MattCheely: I think we can resolve the label decoding issue by not using the `either` decode option and instead using the switch on message type to just decode/validate the `msg` portion of the incoming message.  The switch statement has effectively already checked that the label was valid and known, so all we need to do at that point is decode/validate the message data.

stormojm: I had thought about that after your comments.  It does mean that they could potentially send other properties that aren't being validated.  But I guess we could always do that part manually as well, if we really wanted to block exactness.  And if we did ever expand the root level to include another property it does make it more difficult to validate.

But I guess, those are acceptable compromises.  I can put up the changes tomorrow.

stormojm:
New changes up switch back to using hardcoded strings on the message types.  Then using the decoders to only validate the msg property.

I moved the decoders/validation logic into the individual messages since I didn't like having that msgType hardcoded in multiple spots.  I think the end result is a bit more organized.

js/iframe-coordinator/src/messages/ClientToHost.ts
```diff
@@ -1,57 +1,18 @@
...
 export function validate(msg: any): ClientToHost | null {
-  try {
-    return decoder(msg);
-  } catch {
+  if (!msg || !msg.msgType || !msg.msg) {
     return null;
   }
+
+  return (
+    validateNavRequest(msg) || validatePublication(msg) || validateToast(msg)
```
stormojm: This is a little hard to read if you don't know what the validate methods do.  I considered adding another util for like validateOneOf(vNR, vP, vT) but I don't think that helped a whole lot either...  I am up for suggestions.

js/iframe-coordinator/src/messages/validationUtils.ts
```diff
@@ -0,0 +1,19 @@
+import { Guard } from 'decoders';
+
+export function createMessageValidator<T>(
```
stormojm: Moved a bunch of boilerplate logic here that sprung up once i moved the validates into their own message files.

js/iframe-coordinator/src/messages/validationUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+  messageType: string,
+  decoder: Guard<any>
```
stormojm: using any here is a bit of a trick.  We could also use another Type parameter if we really wanted to make it specific.
But I thought:
createMessageValidator<LabledPublication, Publication>

looked a little off, and the end result was the same anyways.

js/iframe-coordinator/src/messages/ClientToHost.ts
MattCheely: I think it's fine if they're sending extra keys we don't use. That can be a nice tool for version compatibility eventually.

PR #38, opened 2018-10-31
Refactor and simpifying host code.

MattCheely:
This is a partial effort, but it removes the need for two custom
elements, re-uses subscription management logic from the client, and
just reduces the amount of code needed in the host overall. Client
message handling still needs to be added, with verification of incoming
message origin.

@stormojm - Sorry to leave another incomplete PR lying around for you. It shoudn't impact validation, so feel free to ignore it if you don't want to deal with spec failures and whatnot, and I will pick it up again tomorrow. I just want to make sure the direction I'm moving things in is visible.


js/iframe-coordinator/src/HostRouter.ts
```diff
@@ -1,97 +1,73 @@
...
-} from './messages/ClientToHost';
-import {
-  HostToClient,
-  validate as validateOutgoing
-} from './messages/HostToClient';
-import { Publication } from './messages/Publication';
-
 interface ClientRegistration {
   url: string;
   assignedRoute: string;
```
stormojm: I have wondered why we need both the assignedRoute and the key in the RoutingMap.  From what I recall, people would changeRoute using the assignedRoute, so the key was useless.

Could we either changeRoute based of the key and then reuse url for both a full url and Route.  Or just use a straight array for the ClientRegistrations?

js/iframe-coordinator/src/HostRouter.ts
```diff
@@ -1,97 +1,73 @@
...
+        return applyRoute(client.url, route);
+      }
+    });
+    return null;
   }
+}
 
-  public onSendToHost(handler: SubscribeHandler): void {
-    this._toHostSubscriptions.push(handler);
+function matchAndStripPrefix(
```
stormojm: these will need a private or public in front of them.

js/iframe-coordinator/src/HostRouter.ts
MattCheely: Giving each client a unique Id will help us when we integrate things like logging, so there's a logical namespace to log under for each client. It may even be a good idea to tag all messages to the host in general with the id of the client they originate from, so the host can react differently based on the client, if they need to.

js/iframe-coordinator/src/HostRouter.ts
MattCheely: Do they need that since they are not in the class?  I opted to move stateless helpers out of the class and have them not be exposed by the module, in order to make them actually inaccessible to other modules. People shouldn't rely on private fields or those with the underscore prefix, but I think it's better to just keep as much implementation out of the generated objects as possible.

js/iframe-coordinator/src/elements/IframeManager.ts
```diff
@@ -0,0 +1,60 @@
...
+  constructor() {
+    this._frameLocation = 'about:blank';
+
+    this._iframe = document.createElement('iframe');
+    this._iframe.setAttribute('frameborder', '0');
+    this._iframe.setAttribute('style', IFRAME_STYLE);
+    this._iframe.setAttribute('sandbox', 'allow-scripts allow-same-origin');
+  }
+
+  public setFrameLocation(newLocation: string | null) {
```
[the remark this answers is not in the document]
MattCheely: The context here is that it's possible to set the frame location to `null` and get `about:blank`, which is a predictable "no active client" state. We've done that in the past when we've wanted to put an embedded frame in an inactive state without removing it from the DOM. I'm not sure if that use case makes as much sense with this library though.

js/iframe-coordinator/src/HostRouter.ts
stormojm: Makes sense.  As long as there was a purpose behind it. I like the idea of putting in on the events as well.   Probably worth doing once we re-visit the subscription/eventing APIs.

js/iframe-coordinator/src/HostRouter.ts
stormojm: oh, no.  If its out of the class its fine.  I guess i misread the diff.  I am still a noob at github :)

js/iframe-coordinator/src/elements/IframeManager.ts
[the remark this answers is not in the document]
stormojm: I made a typo in my question.  I mean't can we use newLocation?: string instead.  I am fine with having null be sent, but I don't know why we wouldn't allow undefined as well.

js/iframe-coordinator/src/HostRouter.ts
stormojm: One thing I was thinking about while I was adding documentation.  I would like to consider changing changeRoute to use the id instead of the route.  Or have an override which takes that.

At least in PureConnect we will only use urls.  So it would be great if we could just provide the id and the url.  And besides, using an id to change the route seems much more inline with what you had in mind for ids.

js/iframe-coordinator/src/HostRouter.ts
MattCheely: Github diffs are pretty bad.

js/iframe-coordinator/src/FrameManager.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const IFRAME_STYLE =
+  'position: absolute; top: 0; left: 0; width: 100%; height: 100%;';
+
+class FrameManager {
+  private _iframe: HTMLIFrameElement;
+  private _frameLocation: string;
+  private _window: Window;
+
+  constructor(mockWindow?: any) {
```
stormojm: Any reason to not use a configuration object instead of just a mock window parameter?

js/iframe-coordinator/src/FrameManager.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (this._iframe.contentWindow) {
+      const validated = validateIncoming(message);
+      if (validated) {
+        this._iframe.contentWindow.postMessage(validated, clientOrigin);
+      }
+    }
+  }
+
+  public listenToMessages(handler: ((event: ClientToHost) => void)) {
+    this._window.addEventListener('message', event => {
```
stormojm: We will need to removeEventListener at somepoint or we will have a memory leak.

js/iframe-coordinator/src/specs/FrameManager.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+      expect(mocks.mockFrame.contentWindow.postMessage).toHaveBeenCalledWith(
+        message,
+        'http://test.example.com'
+      );
+    });
+  });
+
+  describe('Can be used to subscribe to post messages', () => {
+    it('which are bocked if from a different frame', () => {
+      expect(true).toBe(false);
```
stormojm: looks like this test isn't implemented fully...

MattCheely: This was just a placeholder to remind me where to I left off.

MattCheely:
This is ready for review, and has all of the other typescript migration changes incorporated into it. I will probably squash most of the commits before merging.

js/iframe-coordinator/src/specs/HostRouter.spec.ts
```diff
@@ -1,159 +1,22 @@
 import * as hostRouterInjector from 'inject-loader!../HostRouter';
 import { HostRouter } from '../HostRouter';
-import { LabeledPublication } from '../messages/Publication';
 
 describe('HostRouter', () => {
   let mocks: any;
   let hostRouter: HostRouter;
   beforeEach(() => {
-    /* tslint:disable */
```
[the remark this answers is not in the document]
MattCheely: They should be back now!

js/iframe-coordinator/src/FrameManager.ts
MattCheely: It seemed like just passing the mock window for tests is simpler. I don't think it should introduce any comparability issues if the constructor starts taking arguments in the future, since users of the library shouldn't be passing anything in here. Maybe you're recommending a specific pattern I'm not familiar with though?

js/iframe-coordinator/src/elements/IframeManager.ts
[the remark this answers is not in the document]
MattCheely: Ah, I see. I like the idea of passing `null` because it makes it clear the caller is explicitly and intentionally clearing the location, as opposed to a no-arg call which is a bit more ambiguous, IMO. I'll make the change to allow both though.

js/iframe-coordinator/src/FrameManager.ts
stormojm: I was just thinking if we have to pass 4-5 or more parameter s eventually.  Its better to have your options in a parameter object to reduce having to have loads of parameters to understand.

src/specs/FrameManager.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    /*
+    it('but not if the message is invalid', () => {
+      const message = {
+        msgType: 'notAValidType',
+        msg: {}
+      } as HostToClient;
+      frameManager.sendToClient(message);
+      expect(mocks.frame.contentWindow.postMessage).not.toHaveBeenCalled();
+    });
```
MattCheely: I would like to test this, since it's a use case when called from JavaScript, but I can't get TypeScript to let me pass invalid data in the test!  Any suggestions?

src/elements/frame-router.ts
```diff
@@ -10,10 +16,14 @@ const ROUTE_ATTR = 'route';
  * the client content.
  */
 class FrameRouterElement extends HTMLElement {
```
MattCheely: I would like to add tests for this class, but I haven't been able to inject mocks for dependencies and construct an instance of it in a test successfully (either via `new` or as a custom elment).  In theory I could construct one without intercepting the imports and just monkey-patch the "private" fields, although TypeScript would probably refuse to compile, and it'd be an ugly kludge.

PR #40, opened 2018-11-05
Moving demo-app up a level and fixing an issue with the tests

stormojm:
Reorganize the directory now that elm is no longer being used.


stormojm:
Looks like I made a mistake with my commit messages.  The commit in this moves things under iframe-coordinator sub dir to the root.  It then moves demo-app folder to the root.  It then removes all of our elm code.

PR #41, opened 2018-11-06
First draft of initialization API

stormojm:
The ideal API for me would be if the host automatically sends the data to the client.  However, the client has to be 'listening' before we are allowed to send anything.   This leads us into a bit of a timing chain, since we have to give time for the Client to load the JS, create the Client obj, and call start.  Otherwise, we send data on deaf ears.

That also means the client can't get the data on the initial load of the js, since that has to finish for us to send the data.  At least without sending handshake messages between client/host.

So the approach i took here was to have a callback, that will give the client the data if its set or once it is set.  Its a tad annoying of an API, since we will essentially have to wait for this before we can really do anything with a UI.

The other issue with this approach is that I require start to be called in the initial script load.

The only other thing that could work, afaik, is to have a handshake protocol.  Client would send the ready message once start is called.  Host would respond with another message sending the environment data.

There isn't really a downside to the handshake approach, besides the additional messages and validation we have to do.  So perhaps its the better of the approaches...  Either way we most of the public API's stay the same here.


demo-app/src/index.js
```diff
@@ -52,6 +52,9 @@ router.registerClients(
     return clientMap;
   }, {})
 );
+router.setEnvData({
```
stormojm: The other option is to have this as a part of the registerClients method.  I wasn't sure if anyone had a preference.  Or we could also add attributes to the custom element, but i know working with objects in attributes is a little odd.

justinaray:
I'm doing a bit of a handshake in the setup of the background workers.  The lifecycle is as follows:

* LOADING
    * LOADED
* BOOTSTRAPPING
    * BOOTSTRAP
    * BOOTSTRAP_FAILURE
* RUNNING
    * STOP (Allows graceful shutdown)
* STOPPING

It probably makes sense for all of our clients to have a similar lifecycle.  I'll get the background worker PR up today, so y'all can take a look and we can collaborate.

justinaray:
Here's the PR I mentioned @stormojm:

https://github.com/purecloudlabs/iframe-coordinator/pull/42/files/9997de0ae8269852a95da223d41611e003695866#diff-fe7167fc8c0c6a5fd0e00c7d87846eaeR54

demo-app/public/components/example1/index.html
```diff
@@ -4,6 +4,9 @@
   </head>
   <body>
     <h1>Component #1</h1>
+
+    <h2>Client Data</h2>
+    <textarea id="client-data"></textarea>
```
stormojm: This is just a dumb display for now.  I'll clean this up if we like this API.

demo-app/src/index.js
```diff
@@ -11,22 +11,22 @@ const TOAST_LEVELS = ["info", "success", "error"];
 const TOP_ROUTE_EXTRACTOR = /^#?(\/[^\/]+).*/;
 const NAV_CONFIGS = [
   {
```
stormojm: apparently these include the merge from the parent too... Not sure If i can avoid that with git somehow?

src/FrameManager.ts
```diff
@@ -0,0 +1,138 @@
...
+class FrameManager {
+  private _iframe: HTMLIFrameElement;
+  private _frameLocation: string;
+  private _window: Window;
+  private _postMessageHandler: (event: MessageEvent) => void;
+
+  constructor(options: { onMessage: MessageHandler; mockWindow?: Window }) {
+    this._window = options.mockWindow ? options.mockWindow : window;
+    this._postMessageHandler = event => {
+      this._handlePostMessage(options.onMessage, event);
```
stormojm: I had to change this up a little bit from the parent, since it wasn't actually using the subscription manager to check the interested topics.

src/client.ts
```diff
@@ -80,6 +96,7 @@ class Client {
 
     this._clientWindow.addEventListener('message', this._onWindowMessage);
     this._clientWindow.addEventListener('click', this._onWindowClick);
+    this._sendToHost(Lifecycle.startedMessage);
```
stormojm: trying out a new way to reference messages, since it was really annoying to have the structure here.  Seemed a lot simpler this way.

src/elements/frame-router.ts
```diff
@@ -103,6 +129,33 @@ class FrameRouterElement extends HTMLElement {
...
+      case 'lifecycle':
+        this._handleLifecycleMessage(message.msg as LifecycleStage);
+        break;
+      default:
+        this._dispatchClientMessag(message);
+    }
+  }
+
+  private _handleLifecycleMessage(message: LifecycleStage) {
+    if (Lifecycle.isStartedStage(message)) {
```
stormojm: Another different way to check which message was started.  I liked having the stage name just in one spot, but I am open for suggestions.

src/elements/frame-router.ts
```diff
@@ -103,6 +129,33 @@ class FrameRouterElement extends HTMLElement {
       this.changeRoute(newValue);
     }
   }
+
+  private _handleClientMessage(message: ClientToHost): void {
+    switch (message.msgType) {
+      case 'publish':
+        this._subscriptionManager.dispatchMessage(message.msg);
```
stormojm: I had to change this up from the parent since we weren't actually using subscription manager to determine if we needed to dispatch this message.

demo-app/src/component.js
```diff
@@ -13,7 +13,9 @@ window.onhashchange = function() {
 // Start intercepting link click events for routing
 let iframeClient = new Client();
 iframeClient.start();
-
```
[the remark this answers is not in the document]
MattCheely: Normally I'd say the promise API is the way to go, but we'd have to polyfill for IE11 and I don't want this library to polyfill normally built-in in browser features. I think setting a callback for this specific use case is fine. It should be a one time event, unlike pub-sub type data which can happen many times, so the contract is a little different.

src/messages/Lifecycle.ts
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * A lifecycle message for coordination between host and client.
+ *
+ * @remarks
+ * These are for internal use only.
+ */
+export type LifecycleStage =
+  | LifecycleStarted
+  | LifecycleStopped
+  | LifecycleEnvironmentInit;
```
MattCheely: I'm wondering if we need an explicit life-cycle concept or one this complex. It feels to me like we could get away with just a simple request for environment context and a response from the host. That should be enough for the client to get itself up and running. What other purposes would these life-cycle stages serve?

stormojm: @justinaray had suggested a more complex lifecycle since his background workers used something like that.

I am ok with tuning this back a bit.  However, doing it this way gives us a bit more flexibility for the future.  More than a direct get environment, send environment message would be.

demo-app/src/component.js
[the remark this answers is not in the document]
stormojm: Sounds good to me.  On a related note I think if thats our goal, we may have to look at what our js output is.  Since I believe we are still using Set's which would require a polyfill even with Babel (IIRC).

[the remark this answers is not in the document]
MattCheely: Good call on the use of `Set`. I created a story to follow up on that in general.

src/messages/Lifecycle.ts
MattCheely: I'm on board with structuring the messaging so we could grow to a more complex lifecycle if we need to. I'm just hesitant to build in the concept of a full lifecycle until we know we need it. I think that's one of those concepts that's open-ended enough that we run the risk of folks latching onto it to solve problems that would be better handled in other ways, leading to a lot of coupling between client and host state that will be difficult to maintain. 

For now I think its' good if we have clients send an event to indicate it's ready to start receiving input from the host, but I'd call it something like Ready or Initialized or Started and leave out the Lifecycle bit for now. That way in the future if we really need to, we can grow to something like:

```
export type LifecycleStage = 
    | Ready
    | Stopped
    | ...
```

but we're not building anything conceptually into the API that implies that growth.

stormojm: I think the latest changes should be more in line with what you were thinking.

demo-app/src/index.js
MattCheely: I think making it part of `registerClients` is a good idea, although that might mean re-naming that function to something else. That would let us guarantee that the environment data is in place and ready to go before any clients even start to load. Without that guarantee I can see scenarios where a host and a client could get into weird race conditions where sometimes the environment data would be available and sometimes not.

src/FrameManager.ts
```diff
@@ -73,7 +70,7 @@ class FrameManager {
    */
   public sendToClient(message: HostToClient) {
     const clientOrigin = this._expectedClientOrigin();
-    if (this._iframe.contentWindow) {
+    if (this._iframe.contentWindow && clientOrigin) {
       const validated = validateIncoming(message);
```
MattCheely: Oops. This is a mistake on my part. `validateIncoming` here should be `validateOutgoing`.  I just got lucky that right now the `HostToClient` is a subset of `ClientToHost`

src/FrameManager.ts
```diff
@@ -122,6 +119,10 @@ class FrameManager {
   }
 
   private _expectedClientOrigin(): string {
+    if (this._frameLocation === 'about:blank') {
+      return '';
+    }
```
MattCheely: I think I understand the intent here: don't send messages to `about:blank`.  I considered that restriction, but decided that it was probably unnecessary since `about:blank` should just be an empty page.  I would suggest returning `null` instead of the empty string though. That would make it clearer in the type signatures that `FrameManager` has an expected state which is "No valid origin"

src/messages/Lifecycle.ts
```diff
@@ -0,0 +1,64 @@
...
+  startedDecoder
+);
+
+/**
+ * Environmental data provided to all clients
+ * in order to match behavior of the host application.
+ */
+export interface EnvData {
+  locale: string;
+}
```
MattCheely: I would add `hostRootUrl` here. There's a lot of useful information that can be derived from the client knowing who's hosting it. Also, it will enable the client to generate links that work correctly when opened in a new tab or shared via email/chat in the cases where URL routing is used.

demo-app/src/component.js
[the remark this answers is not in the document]
MattCheely: > Normally I'd say the promise API is the way to go, but we'd have to polyfill for IE11 and I don't want this library to polyfill normally built-in in browser features. I think setting a callback for this specific use case is fine. It should be a one time event, unlike pub-sub type data which can happen many times, so the contract is a little different.

:point_up: I'm reconsidering my thoughts on emitting this as an event after looking at the pub-sub API proposal. See my comments there for more.

src/FrameManager.ts
stormojm: There was some case where I ran into this... I think it was on the demo app?  It was throwing an error which i figured we didn't want to do.  I'll change it to null and make sure thats anticpated.

stormojm:
I rebased this PR off of the pub/sub API so we can get the getEnvData using the same approach.  Therefore some of the issues from that PR will leak through to here, but I'll fix that in that branch.

Besides that I think all the comments are accounted for.

src/client.ts
```diff
@@ -19,12 +24,18 @@ interface ClientConfigOptions {
   clientWindow?: Window;
 }
 
+/**
+ * A strictly-typed event handler for publication messages.
+ */
+type ClientEventEmitter = EnvDataEventEmitter & PublicationEventEmitter;
```
stormojm: Most of this will change with the changes to PubSub, but I wanted to get the init api flowing through events.

src/FrameManager.ts
MattCheely: Oh, nice catch. I didn't expect it to throw errors when on about:blank. Shows what I know.

src/messages/Lifecycle.ts
MattCheely: I think this is an API that would benefit from a `custom: any` field that could pass through arbitrary application-specific data. That's actually where I'd handle `platformId`, for a couple of reasons. One is that I'd like to keep the library geared toward general-purpose use and not encode too much that's specific to Genesys into it. The other is that we may end up using this in some unexpected cases for things like internal tooling where `platformId` doesn't make sense, or would end up growing in a way that might not be useful. 

What's the logical difference between locale and language in this case?  In PureCloud we don't separate the two, so I suspect this is a feature difference between platforms that I'm just not expecting.

stormojm: I'll update to use a custom.  Of course we won't be able to validate those data, so I would probably discourage using that as much as possible since its an implicit API.

Language is used to determine the translations the UI strings use.  Locale controls how certain things are formatted (numbers [3,2 vs 3.2], currency, and datetimes).  It also is supposed to control how things are sorted, but that doesn't work great in JS yet.
Separating these two allows you to write fewer translation packs while still working in many regions (de-DE vs de-AT for example).

Also in theory you might be forced to use an english translation, but be located in another country (e.g. Netherlands) and wish to display the values in the location.

justinaray: As @MattCheely mentioned, we usually just let the user select their preferred language and that is represented in a single value, `userPrefLangTag`.  This is in the format of a BCP 47 lang tag.  This allows us to pick the best supported language bundle we have for the user (dropping dialects) and also allows fallback based on the user agent's language property which is also in BCP 47.  Also, bcp 47 format is supported by the JS intl api.

Perhaps, we should be more specfic either in property name or type to indicate which type of value is expected here.  We've got a lot of ugly legacy code transforming all manor of language identifiers from our earlier days :(

+1 on separating language from other concerns.  I'm just not sure locale goes far enough or is the most obvious label.  Perhaps we could be more specific?  Something like dateFormatLangTag, numberFormatLangTag, phoneNumberLangTag?  All of these could fall back to the default langTag if not specified.  At some point, we'll have to cut it off and move into the custom section.

stormojm: In PureConnect, we use separate codes.  We default their display language to the browser language if we can as well, but that may not always be the choice they want to use for the displayed language (there may be a native Spanish speaker in an english call center).  And we always use their browser setting for their locale (used with JS Intl and a couple of our third-party dists).

Has anyone talked with Localization what their plans are?  That would certainly help us decide this I think.

Also I really don't think we need to have more specific tags.  It's extremely rare that you would use one locale for numbers and another for dates. As for naming... .NET calls these CommonCulture and CommonUICulture (https://blogs.msdn.microsoft.com/heaths/2005/02/17/the-differences-between-locales-and-languages/).  In this draft WEC document they used 'Langauge Tag' and 'Locale Identifier': https://www.w3.org/TR/ltli/.

MattCheely: Here are the options I see for language/locale stuff:

1. Keep it as is with `language` and `locale`. That lets us handle scenarios where language is different from locale, which is a nice feature if the platform supports it. Where they're the same, we can just use the same value. It doesn't provide for more complex customization options, although at that point, I think we'd want apps to make an API call or use the `custom` field anyway. I do worry a little bit that in platforms where they're not different, it's not clear which one is the source of truth (they shouldn't be different, but mistakes happen).
2. Just have `locale` since it's the _minimum_ necessary to render, and defer any additional options to `custom` on a per-platform basis.
3. Decide the problem isn't well understood right now, take them both out, and rely on `custom` until some consistent pattern emerges.

Personally, I prefer options 2 and 3 but the API as it is now isn't a deal-breaker, IMO.

PR #42, opened 2018-11-06
Feature/pcui 405 background workers

justinaray:
I still have a few todos and some clean-up to do on this PR, but I wanted to get it up for discussion on:

* The Lifecycle API (As it relates to Init)
* Where we store our types
* How we'll manage message types and validation in the ts version rather than elm


js/demo-app/src/index.js
```diff
@@ -41,17 +41,37 @@ registerElements();
...
       assignedRoute
     }
 
     return clientMap;
-  }, {})
-);
+  }, {}),
+  ...{
+    // Standard worker
+    'backgroundWorker1': {
```
justinaray: This is the first shot at the API for registering a background worker.  Just a simple boolean and a url is all that's necessary.  These types are tracked in typescript as a union type, so I didn't add a type property at this point, however, I could easily switch to that.  Thoughts?

js/demo-app/src/workers/demo-worker.js
```diff
@@ -0,0 +1,24 @@
+import BackgroundClient from "iframe-coordinator/BackgroundClient";
+// TODO Need a better demo that makes more sense?
+
+const client = new BackgroundClient();
```
justinaray: Took a fair amount of webpack finagling to get this to bundle correctly as a worker.  But the end result is cool!  All the messaging infrastructure is hidden behind the BackgroundClient api.

js/iframe-coordinator/libs/BackgroundClient.ts
```diff
@@ -0,0 +1,35 @@
...
+    { title = null, custom = null }: ToastOptions = {}
+  ): void {
+    this._publishMessageToHost(WorkerToHostMessageTypes.ToastRequest, {
+      title,
+      message,
+      custom
+    });
+  }
+
+  // TODO Change this to URL based?
```
justinaray: @MattCheely - we talked about converting this to a straight URL.  Should I keep it as a route for now?

js/iframe-coordinator/libs/BackgroundClient.ts
```diff
@@ -0,0 +1,35 @@
...
+  public requestNavigation(route: string): void {
+    this._publishMessageToHost(WorkerToHostMessageTypes.NavRequest, {
+      fragment: route
+    });
+  }
+
+  private _publishMessageToHost(msgType: string, msg?: object): void {
+    ctx.postMessage({
+      protocol: WORKER_MESSAGING_PROTOCOL_NAME,
```
justinaray: Having worked with a lot of messaging between iframes, we found that having a protocol field with a specific string is a great way to prune out invalid or unintentional messages.  Other devs are not always careful to domain-restrict or target their messages appropriately.

I also usually add an agent name and version so we can track api clients, but that can come later.  I think we should adapt this pattern to all of our messages

js/iframe-coordinator/libs/Elm.d.ts
```diff
@@ -46,7 +46,21 @@ interface HostProgram {
...
+  url: string;
+  assignedRoute: string;
+}
+
+interface BackgroundClientConfig {
+  background: boolean;
+  url: string;
+}
+
+type ClientConfig = RoutableClientConfig | BackgroundClientConfig;
```
justinaray: Here's the union type to allow for configuration of different types.  I can handle this with a more explicit type if we decide that's better.

js/iframe-coordinator/libs/client.ts
```diff
@@ -1,16 +1,11 @@
...
-  custom?: any;
-}
+import { PublicationHandler, ToastingClient, ToastOptions } from './types';
 
 interface ClientConfigOptions {
   clientWindow?: Window;
 }
 
-class Client {
+class Client implements ToastingClient {
```
justinaray: Pulling in interfaces for features that clients can implement will keep the APIs consisent for client types (Traditional, Background, future types).

If we like this approach, I will add the NavigatingClient interface as well.

js/iframe-coordinator/libs/workers/worker-manager.ts
```diff
@@ -0,0 +1,288 @@
...
+ * support lenient shutdown?
+ *  notification on shutdown?
+ *  soft stop
+ *  allow restart?
+ * support client stop request?
+ *   worker never can close now as i have attached listeners
+ *   but, you can't re-add now either via registerclients
+ */
+
+enum WorkerPhase {
```
justinaray: @stormojm ... here's the worker lifecycle I mentioned in the other PR.  I would be 100% on board with elevating this to a ClientLifecycle so we have consistent phases and events to advance the client through.  It could also be extended if clients have different phases, etc.s

js/iframe-coordinator/libs/workers/worker-manager.ts
```diff
@@ -0,0 +1,288 @@
...
+    this._workers = [];
+  }
+
+  public start(url: string): string {
+    const id = `${WorkerManager.ID_PREFIX}${++WorkerManager._workerIndex}`;
+
+    // @ts-ignore
+    const worker = new SpawnWorker();
+    WorkerManager.trackedWorkerEventTypes.forEach(curr => {
+      worker.addEventListener(curr, this);
```
justinaray: I call this out just because I thought it was cool and I didn't previously know you could attach an object implementing `EventListenerObject`.

This is cool because it makes it trivial to remove listeners for clean-up.  No need to track function references!

js/iframe-coordinator/libs/workers/worker-manager.ts
```diff
@@ -0,0 +1,288 @@
...
+      WorkerManager.trackedWorkerEventTypes.forEach(curr => {
+        worker.removeEventListener(curr, this);
+      });
+      worker.terminate();
+
+      this._workers.splice(workerIndexToStop, 1);
+    }
+  }
+
+  // TODO Need comms to workers (pub/sub, others?)
```
justinaray: @MattCheely Should background workers be able to subscribe/publish to any channel?  Should the host app be able to "authorize" or intercept and `preventDefault`?

js/iframe-coordinator/libs/workers/worker-manager.ts
```diff
@@ -0,0 +1,288 @@
...
+            // tslint:disable-next-line
+            console.error('Error rate exceeded.  Stopping the worker', {
+              workerDetails: managedWorker,
+              errorWindowCount: managedWorker.errorWindowCount,
+              errorWindowStartTimestamp:
+                managedWorker.errorWindowStartTimestamp,
+              errorWindowDurationMillis: delta,
+              errorDetails: evt.error
+            });
+            this.stop(managedWorker.id);
```
justinaray: I decided on a configurable error rate.  If exceeded, we will log and stop the worker.  We can tune this as needed.

Also, we could allow the client to configure a restart or allow the register function to be recalled.  Thoughts?

js/iframe-coordinator/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "module": "commonjs",
     "declaration": true,
     "declarationDir": "./",
-    "target": "es5"
+    "target": "ES6",
```
justinaray: I bumped this up to allow the usage of some functional array methods.

@MattCheely we probably want to distribute es6 modules (ts and/or js) and es5 UMD modules for maximum consumability and tooling, correct?

js/iframe-coordinator/webpack.config.js
```diff
@@ -20,6 +20,17 @@ module.exports = {
           options: { debug: true }
         }
       },
+      {
+        test: /\.worker\.(js|ts)$/,
+        exclude: [/node_modules/],
+        use: {
+          loader: "worker-loader",
```
justinaray: This is a pretty awesome loader.  Allows a web-worker to be included into the baseline via a BLOB and avoids the network call and need to ask library users to deploy a worker along with our library.

js/demo-app/src/workers/demo-worker.js
stormojm: Can we make this a non-default export?  It would match our other entry points and let us potentially expand the exports eaiser.

js/iframe-coordinator/libs/BackgroundClient.ts
stormojm: how is this different than the msgType?  It's really just acting like a combined key with the msg type.  I guess if you have had issues before, i'll defer.  Just seems extra cautious.

js/iframe-coordinator/libs/Elm.d.ts
stormojm: If we only provide a background client config, we will still work ok, right?  If so I think this is fine.

js/iframe-coordinator/libs/elements/frame-router.ts
```diff
@@ -30,13 +35,57 @@ class FrameRouterElement extends HTMLElement {
   }
 
   public registerClients(clients: ClientRegistrations) {
```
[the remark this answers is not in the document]
stormojm: If we do allow dynamic, I wonder if its better to have just add/removeClients methods.

js/iframe-coordinator/libs/BackgroundClient.ts
justinaray: The issue is that a lot of sites are using some form of iframe/postMessage these days and a growing number of pages and frameworks are using workers.  And, I've seen a fair share that don't do a good job targeting their messages to a particular frame or adding a targetOrigin.  

Having an unambiguous field like protocol and giving it a unique value allows us to effectively filter in our code.  In other projects, I had a ton of false-positive error logging because I was processing messages (and logging warnings of malformed payloads) from embedded urls that were pushing events over postMessage.  They all use generic fields like type, msgType, etc.

Now, we could get a similar result by namespacing our msgType values.  Something like `com.genesys:iframe-coordinator:requestToast` would allow us to filter effectively and also reduces the likelihood that our messages would be matched and consumed by others.

So, I could certainly go this way as well if we'd rather not add an extra field.

js/iframe-coordinator/libs/elements/frame-router.ts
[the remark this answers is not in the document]
justinaray: Yeah, I would definitely agree we'd change to and `add/remove` paradigm if we go this route.

js/iframe-coordinator/tsconfig.json
stormojm: Are there still places where people don't transpile?  It would be nice if we could only distrubute es6.

js/iframe-coordinator/libs/BackgroundClient.ts
stormojm: I think i prefer a separate field so the msgType becomes more meaningful/easier to work with.

MattCheely:
Man, when we stubbed out this behavior, I was pretty sure somebody was going to come along and need it soon, and it never happened. I should have closed this PR some time ago, but I will still keep the branch around in case this comes back in some for.

PR #45, opened 2018-11-14
Better pub/sub API?

stormojm:
Still a work in progress, but I wanted to put this up in case anyone wanted to give feedback on this API.

### DIFF
--- a/demo-app/src/component.js
+++ b/demo-app/src/component.js
@@ -12,11 +12,15 @@ window.onhashchange = function() {
 
 // Start intercepting link click events for routing
 let iframeClient = new Client();
-iframeClient.start();
+iframeClient.addListener("environmentalData", envData => {
+  document.getElementById("client-data").value = JSON.stringify(envData);
+});
 
 iframeClient.messaging.addListener("host.topic", publication => {
   console.log("Got Publish event:", publication);
 });
+
+iframeClient.start();
 
 document.getElementById("do-publish").addEventListener("click", () => {
   //Get the topic and the data to publish from the form
--- a/demo-app/src/component.js
+++ b/demo-app/src/component.js
@@ -14,9 +14,10 @@ window.onhashchange = function() {
 let iframeClient = new Client();
 iframeClient.start();
 
-iframeClient.onPubsub(publication => {
+iframeClient.on('publish', publication => {
   console.log("Got Publish event:", publication);
 });
+
 iframeClient.subscribe("host.topic");
 
 document.getElementById("do-publish").addEventListener("click", () => {
--- a/demo-app/src/component.js
+++ b/demo-app/src/component.js
@@ -14,7 +14,7 @@ window.onhashchange = function() {
 let iframeClient = new Client();
 iframeClient.start();
 
-iframeClient.on("host.topic", publication => {
+iframeClient.messaging.addListener("host.topic", publication => {
   console.log("Got Publish event:", publication);
 });
 
--- a/demo-app/src/component.js
+++ b/demo-app/src/component.js
@@ -14,11 +14,9 @@ window.onhashchange = function() {
 let iframeClient = new Client();
 iframeClient.start();
 
-iframeClient.on('publish', publication => {
+iframeClient.on("host.topic", publication => {
   console.log("Got Publish event:", publication);
 });
-
-iframeClient.subscribe("host.topic");
 
 document.getElementById("do-publish").addEventListener("click", () => {
   //Get the topic and the data to publish from the form
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
   "dependencies": {
     "@types/jasmine": "^2.8.9",
     "decoders": "^1.10.6",
-    "events": "^3.0.0",
     "request": "^2.88.0"
   },
   "files": [
--- a/package.json
+++ b/package.json
@@ -5,9 +5,9 @@
   "dependencies": {
     "@types/jasmine": "^2.8.9",
     "decoders": "^1.10.6",
-    "elm": "^0.19.0-bugfix2",
-    "eventemitter3": "^3.1.0",
-    "request": "^2.88.0"
+    "events": "^3.0.0",
+    "request": "^2.88.0",
+    "strict-event-emitter-types": "^2.0.0"
   },
   "files": [
     "client.js",
--- a/package.json
+++ b/package.json
@@ -6,8 +6,7 @@
     "@types/jasmine": "^2.8.9",
     "decoders": "^1.10.6",
     "events": "^3.0.0",
-    "request": "^2.88.0",
-    "strict-event-emitter-types": "^2.0.0"
+    "request": "^2.88.0"
   },
   "files": [
     "client.js",
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "@types/jasmine": "^2.8.9",
     "decoders": "^1.10.6",
     "elm": "^0.19.0-bugfix2",
+    "eventemitter3": "^3.1.0",
     "request": "^2.88.0"
   },
   "files": [
--- a/src/EventEmitter.ts
+++ b/src/EventEmitter.ts
@@ -1,0 +1,177 @@
+/**
+ * A generic handler for incoming event data.
+ */
+type THandler<T> = (data: T) => void;
+
+/**
+ * Simple object for keeping track of event listeners.
+ */
+interface Events<T> {
+  [index: string]: Array<THandler<T>>;
+}
+
+const findIndex =
+  [].findIndex ||
+  // IE11 support
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Polyfill
+  function(
+    predicate: (value: never, index: number, obj: never[]) => boolean,
+    thisArg?: any
+  ) {
+    if (this == null) {
+      throw new TypeError('"this" is null or not defined');
+    }
+
+    const o = Object(this);
+    // tslint:disable-next-line
+    const len = o.length >>> 0;
+
+    if (typeof predicate !== 'function') {
+      throw new TypeError('predicate must be a function');
+    }
+
+    let k = 0;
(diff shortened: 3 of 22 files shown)


src/client.ts
```diff
@@ -1,3 +1,6 @@
+import * as EventEmitter from 'events';
+import StrictEventEmitter from 'strict-event-emitter-types';
```
stormojm: There are a few alternatives to this, but they are all reimplementations of EventEmitter.  Using strict-event-emitter uses fancy typescript (https://medium.com/@bterlson/strongly-typed-event-emitters-2c2345801de8) with the implementation backed by popular events (or eventemitter3) package for the actual eventing.

src/client.ts
```diff
@@ -19,17 +22,38 @@ interface ClientConfigOptions {
   clientWindow?: Window;
 }
 
+/**
+ * The known events that can be emitted
```
stormojm: This is a little annoying because this is another definition of the ClientToHost messages we have.

demo-app/src/component.js
```diff
@@ -14,9 +14,10 @@ window.onhashchange = function() {
 let iframeClient = new Client();
 iframeClient.start();
 
-iframeClient.onPubsub(publication => {
+iframeClient.on('publish', publication => {
   console.log("Got Publish event:", publication);
 });
+
 iframeClient.subscribe("host.topic");
```
stormojm: This is one of my bigger open questions for API design. 

With this approach, we would have to still subscribe to publish topics.  But it provides a common API for any event that we wanted to publish.

Alternatively we could say that the on will only emit Publish events.
Thus our api would be `.on('host.topic', handler)`

Now that I write this, I think that might be better...

src/client.ts
```diff
@@ -165,4 +183,4 @@ class Client {
   }
 }
 
-export { Client };
+export { Client, Publication };
```
stormojm: TS consumers will need to know what a publication is to provide the right handler.

src/client.ts
stormojm: Removed this library in favor of our own interface definition.  This allows us to have a bit more control over the types of emits/on's that we provide.

demo-app/src/component.js
stormojm: Updated the API to `.on('host.topic', handler)` style API.

src/messages/Publication.ts
```diff
@@ -14,6 +14,24 @@ export interface Publication {
...
 
+/**
+ * Handles publications
+ */
+export type PublicationHandler = (publication: Publication) => void;
+
+/**
+ * A strictly-typed event handler for publication messages.
+ */
+export declare interface PublicationEventEmitter {
```
stormojm: As we add more HostToClient messages (that need to notify the client) we would have to expand this interface using & to include WhateverEventEmitter).

MattCheely:
I like the direction this is heading.  It's made me think a bit more about similarities between the client and host APIs for inbound messages to be handled by application code.  On the host, those are already a sort of event emitter because we use custom events. That could make sense on the client as well. Publication is a bit odd because it's really just nesting another general-purpose set of events/messages in the higher-level API, so it makes sense for it to have it's own event emitter centered around topics.  

I'm thinking we're headed toward an API like this for inbound events:

```
// Implement now for PubSub
client.pubSub.addListener(pubSubTopic, pubSubHandler)
client.pubSub.removeListener(...)
client.pubSub.removeAllListeners(...)

// Implement for environmentData on load and future events
client.addListener('envrionmentData', someMessageHandler)
client.removeListener(...)
client.removeAllListeners(...)

// frame-router element PubSub
frameRouterElement.pubSub.addListener(pubSubTopic, pubSubHandler)
frameRouterElement.pubSub.removeListener(...)
frameRouterElement.pubSub.removeAllListeners(...)

// frame-router primary api events
frameRouterElement.addEventListener('toastRequest', toastHandler)
//... all of the built in DOM event handling APIs ...
```

Thoughts?

src/client.ts
```diff
@@ -22,14 +22,13 @@ interface ClientConfigOptions {
 /**
  * The Client is access point for the embedded UI's in the host application.
  */
-class Client {
-  private _subscriptionManager: SubscriptionManager;
+class Client extends (EventEmitter as { new (): PublicationEventEmitter }) {
```
MattCheely: I generally try to opt for composition over extension/inheritance. I think this is a good case where instead of extending `EventEmitter`, we would benefit from creating a new `EventEmitter` as a private property of the class, and delegating calls to it.  That lets us expose less of it's API (see other comments) and be less coupled to a specific implementation.

src/messages/Publication.ts
MattCheely: This feels like a bit much in terms of a public-facing API. 

I'd suggest that:
 - `emit` should be private from the application perspective. We don't want applications using the libraries as a general purpose eventing bus. 
- `on` and `addListener` are redundant. We should expose `on`/`off` or `add...`/`remove...` and leave it at that.
- `once` isn't really necessary and opens up a can of worms about how many eventListener helper functions do we add. I'd rather folks use other libraries and tools that can provide that functionality on top of a basic event handler
- `listenerCount` doesn't seem very useful externally.

package.json
```diff
@@ -5,7 +5,7 @@
   "dependencies": {
     "@types/jasmine": "^2.8.9",
     "decoders": "^1.10.6",
-    "elm": "^0.19.0-bugfix2",
+    "events": "^3.0.0",
```
MattCheely: I do have a concern about the events package, which is this in the docs for `on` (emphasis mine):

> Adds the listener function to the end of the listeners array for the event named eventName. **No checks are made to see if the listener has already been added. Multiple calls passing the same combination of eventName and listener will result in the listener being added, and called, multiple times.**

It don't know why node opted to create a non-idempotent event API, but it seems a bit crazy to me. It makes it much harder to work with as a client. It's also not really expected in the browser space where [element event listeners behave differently](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Multiple_identical_event_listeners):

> If multiple identical EventListeners are registered on the same EventTarget with the same parameters, the duplicate instances are discarded. They do not cause the EventListener to be called twice, and they do not need to be removed manually with the

src/messages/Publication.ts
stormojm: Yea these are fair points.  I just grabbed everything form EventEmitter without really thinking about it.

Just a point of clarification though.  JS users will still be able to get these if we expose the EventEmitter object (or extend from it).  These are just limitations for TS users.  I could always add a Facade though, if we want to restrict JS as well?

stormojm: NM, i read your comment down below.  I'll change that.

stormojm:
> I like the direction this is heading. It's made me think a bit more about similarities between the client and host APIs for inbound messages to be handled by application code. On the host, those are already a sort of event emitter because we use custom events. That could make sense on the client as well. Publication is a bit odd because it's really just nesting another general-purpose set of events/messages in the higher-level API, so it makes sense for it to have it's own event emitter centered around topics.
> 
> I'm thinking we're headed toward an API like this for inbound events:
> 
> ```
> // Implement now for PubSub
> client.pubSub.addListener(pubSubTopic, pubSubHandler)
> client.pubSub.removeListener(...)
> client.pubSub.removeAllListeners(...)
> 
> // Implement for environmentData on load and future events
> client.addListener('envrionmentData', someMessageHandler)
> client.removeListener(...)
> client.removeAllListeners(...)
> 
> // frame-router element PubSub
> frameRouterElement.pubSub.addListener(pubSubTopic, pubSubHandler)
> frameRouterElement.pubSub.removeListener(...)
> frameRouterElement.pubSub.removeAllListeners(...)
> 
> // frame-router primary api events
> frameRouterElement.addEventListener('toastRequest', toastHandler)
> //... all of the built in DOM event handling APIs ...
> ```
> Thoughts?

One issue with this approach is that there would be no way for the client to request the en

demo-app/src/component.js
```diff
@@ -14,10 +14,9 @@ window.onhashchange = function() {
 let iframeClient = new Client();
 iframeClient.start();
 
-iframeClient.onPubsub(publication => {
+iframeClient.messaging.addListener("host.topic", publication => {
```
stormojm: I used messaging instead of pubSub, just because I think its friendlier... Open for feedback

src/EventEmitter.ts
```diff
@@ -0,0 +1,177 @@
...
+function isRegistered(value: THandler<any>) {
+  return value === this;
+}
+
+/**
+ * Exposes a limited subset of event emitter
+ * functionality to ensure external modules can
+ * not self-emit.
+ */
+export class ExposedEventEmitter<T> {
```
stormojm: Is there is a better way to make it so JS people can't access dispatch?

src/client.ts
```diff
@@ -82,6 +89,13 @@ class Client {
     this._clientWindow.addEventListener('click', this._onWindowClick);
   }
 
+  /**
+   * Eventing for published messages from the host application.
+   */
+  public get messaging(): ExposedEventEmitter<Publication> {
```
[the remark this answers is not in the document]
MattCheely: I think that makes more sense. Good call.

src/EventEmitter.ts
MattCheely: Sadly, I don't know of a a better option. The delegation would be more concise in vanilla JS, but the structure would be pretty much the same.

src/EventEmitter.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Simple object for keeping track of event listeners.
+ */
+interface Events<T> {
+  [index: string]: Array<THandler<T>>;
+}
+
+const findIndex =
+  [].findIndex ||
```
MattCheely: I know we said earlier that `Set` wasn't an option for IE11, but the [MDN compatiability table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#Browser_compatibility) makes it seem like it's compatible enough.

PR #46, opened 2018-11-15
Move build scripts to npm ci instead of install

justinaray:
This ensures the lockfile is used for standard runs and lets devs update the lockfile as needed through standard npm commands.

### DIFF
--- a/scripts/buildall.sh
+++ b/scripts/buildall.sh
@@ -6,5 +6,5 @@ ${PROJ_DIR}/scripts/buildlibs.sh
 
 echo "** Building & running demo-app **"
 cd ${PROJ_DIR}/demo-app
-npm install
+npm ci 
 npm start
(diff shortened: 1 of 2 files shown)


scripts/buildall.sh
```diff
@@ -6,5 +6,5 @@ ${PROJ_DIR}/scripts/buildlibs.sh
 
 echo "** Building & running demo-app **"
 cd ${PROJ_DIR}/demo-app
-npm install
+npm ci
```
justinaray: This way, package.lock is used for local build/serve and for ci builds.  We could certainly parameterize this, but I feel pretty strongly that the lockfile shouldn't be updated unless you also update the package.json file.  i.e. you're intentionally updating dependencies.  Feel like this is a pretty good default and devs in the know can manually run the npm commands they want to bump.

The other approach is to not use semver ranges in package.json, which seems not great either.

Feel free to weigh in here with opinions.  The community certainly has as well.

PR #47, opened 2018-11-20
Better documentation for IE11 support

stormojm:
This does two things:
1. Documents the polyfills we require for IE11 (@justinaray had already added them)
2. Document that transpiling is required.

Our story for IE will thus be up to the Host and Client apps to handle appropriately.  If there is anything more I should add, let me know.


README.md
```diff
@@ -102,3 +110,5 @@ client.start();
     <!-- ... -->
 </body>
 ```
+### IE11 support
+Our target version of javascript is ES2015.  This means that you will be required to transpile this library if you wish to support IE11.  In addition the necessary polyfills will need to be loaded by both the host application and the client frame.
```
justinaray: +1 for this.  I think not transpiling or including polyfills allows for maximum consumption avenues.

I think it would ultimately be cool to also compile our ts files, but output the assests as js es modules.  That allows for direct consumption of the modules by shops fully embracing es modules and a tooled build.

Internally, I think we could consume this library and build a UMD es5 version with transpiled code and polyfills for our supported browsers that could be used in require or CDN model if desired.

README.md
```diff
@@ -15,6 +15,14 @@ Embedding applications via iframe also means that code from separate teams can b
 **JavaScript**
 
 ```js
+/*
+ * We require a few polyfills in order to support IE11.  These
+ * will be needed to be loaded by both the host and the client.
+ */
+import "@babel/polyfill";
+import "custom-event-polyfill/polyfill.js";
+import "url-polyfill";
```
justinaray: Thanks for documenting these!

justinaray: Should we add anything about `browserslist` and `preset-env` or maybe just a link on how to use the babel polyfills?

demo-app/package.json
```diff
@@ -39,7 +39,6 @@
     "@babel/polyfill": "^7.0.0",
     "@webcomponents/custom-elements": "^1.0.9",
     "copy-webpack-plugin": "^4.5.2",
-    "core-js": "^2.5.7",
```
justinaray: Trying to remember my rationale on why I directly added this.  We certainly get it transitively from @babel/polyfill.  Maybe I was thinking it would be obvious that we can use es6 features in this baseline?

Assuming it will come in fine transitively.  Did it run on ie11 ok?

stormojm: Yeap, everything works in IE still

README.md
stormojm: I want to be careful about overspecifing too much.  Since the host app probably has lots of concerns.  Heck even babel polyfill itself probably not required (We have a different one in PureConnect for example).

But I am up for adding anything we feel may be helpful

PR #48, opened 2018-11-21
Client id messages on each request

stormojm:
Adding a message for passing client id to the client, which can then provide the client id on each request


src/HostRouter.ts
```diff
@@ -29,6 +29,23 @@ export class HostRouter {
 
     return clientUrl;
   }
+
+  /**
+   * Gets the client id for a provided route.
+   *
+   * @param route The route to lookup, such as '/foo/bar/baz'
+   */
+  public getClientId(route: string): string | null {
```
stormojm: I would like to merge this into getClientUrl before pushing.

src/elements/frame-router.ts
```diff
@@ -126,6 +128,13 @@ class FrameRouterElement extends HTMLElement {
   }
 
   private _handleLifecycleMessage(message: LabeledStarted) {
+    this._frameManager.sendToClient({
```
stormojm: This sends a message to the client with the id that the host thinks it is.
I used a confirmationId since we could in theory have two clients going through initialization, so its just a way to make sure that it confirms that it got the right client.

Although, this still doesn't solve the issue fully since the host doesn't know which temp id is which...  Maybe this situation isn't worth programming around?

Alternative to this is that we could have the temp id be the id passed from the client be permanent.  Then the FrameRouter can reconcile the random id to the configured id.

src/messages/Publication.ts
```diff
@@ -21,22 +21,34 @@ export type PublicationHandler = (publication: Publication) => void;
...
 export interface LabeledPublication extends LabeledMsg {
   msgType: 'publish';
   msg: Publication;
 }
 
+/**
+ * A message used to publish a generic messages
+ * to the host application.
+ */
+export interface LabeledClientPublication extends LabeledPublication {
```
stormojm: Can no longer reuse the same interface for host-to-client and client-to-host, unless we wanted to make the client id into sender id.

src/client.ts
```diff
@@ -177,6 +189,7 @@ class Client {
   public publish(publication: Publication): void {
     this._sendToHost({
       msgType: 'publish',
+      clientId: this._clientId,
```
justinaray: I will add a similar id to my worker bootstrap lifecycle method.

@MattCheely and I were discussing this.  This will allow the host app to make decisions on priority and honoring the request (nav, toast, etc.) based on what client is requesting and what's currently running.

src/genTempId.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Generates a temporary id which is semi-unique.
+ */
+export function genTempId(): string {
+  return String(Math.floor((1 + Math.random()) * 0x10000));
```
justinaray: my pr's bringing in a uuid lib capabile of generating uuidv4 guids if you want to use it.

src/elements/frame-router.ts
justinaray: Any reason to not include this in env init message?

Also, it occurs to me now that maybe we just send the id used to register the clients?  So that the host app can make the decisions without an opaque key?  I guess there's a chance where the host doesn't want the client to have it's id, but that seems unlikely and not worth the trouble of mapping between the key spaces.

justinaray: I saw this, thought I initially understood it, commented on it, and now I'm starting to question it all 😱

I was thinking of lots of edge cases.  The one you pointed out that an init procedure can be interrupted at any time by a user navigating to another client that results in a page load.  Also, there's the case where the frame is using hash routing internally, but not playing nice with our routing and navigating between the pages.

As such, I'm wondering if it even makes sense to try and exchange IDs or track them.  Maybe, the best approach is to just let the message come in and be decorate userspace events (events handled by the host embedding app) with the current client id and url.  We could also dump the current client id if the frame has been navigated and doesn't match the expected origin.

In this way, user land can still know the client id of the (toast, nav, etc.) requestor, and we dont have to worry about state and race conditions.

Suspect we'll have to chat a bit more about this, but just wanted to get some thoughts out.

MattCheely: I'm not sure the client needs to be in the loop about what it's id is. We also don't want the host to have to trust the client apps to not send incorrect identifiers.  The `FrameManager` class already knows what domain it expects the client to be on and doesn't allow any messages that aren't from that domain. We should be able to track the expected client id on the host side, and add it there.

If we want additional protections against accidentally mixing up same-domain clients, we could even validate the current location of the client window against the expected client path when tagging the message with the client id.  Technically it's possible that the client frame navigated after a valid message was sent and we could drop messages that should be valid, but the client frames aren't supposed to navigate on their own anyway.  It shouldn't be an issue for a client that is behaving correctly.

stormojm: I dropped the client_id passing to the client.  Now frame-router will add the "current client id" to the messages.

src/elements/frame-router.ts
```diff
@@ -133,8 +137,11 @@ class FrameRouterElement extends HTMLElement {
   }
 
   private _dispatchClientMessage(message: ClientToHost) {
+    const messageDetail: any = message.msg;
+    messageDetail.origin = this._currentClientId;
```
MattCheely: I think I liked `clientId` better than origin. I'm a bit worried about the potential for confusion with the existing meaning of origin in the contest of domains.

src/HostRouter.ts
MattCheely: Yeah, I think something like getClientInfo makes more sense once we start doing both.

MattCheely:
I added a couple of comments, but I think we're good to go after those. Thanks, @stormojm!

src/elements/frame-router.ts
stormojm: Yea, I had debated about the same thing.  I ended up this way just because origin sounds nicer, but I'll change it back to avoid the confusion.

src/HostRouter.ts
stormojm: I changed this around.  Although it was a tad confusing, since ClientInfo is the type used by the original maps.

I considered returning a ClientInfo, but the assignedRoute and url were both different from the original.  So i thought that would be confusing.  Instead I went with a non-defined type with just the id/url.  Its slightly ugly, but I think its more clear from a consumer perspective.

src/elements/frame-router.ts
MattCheely: There's clearly an insufficient supply of not-already-taken words in programming. Maybe we should all learn Japanese or French so we can get a whole new free set to use.

PR #50, opened 2018-11-30
Handle a few invalid origin edge cases

justinaray:
Came across a few edge cases while working on something else.  Figured I'd add them in a sep PR.

### DIFF
--- a/src/FrameManager.ts
+++ b/src/FrameManager.ts
@@ -106,8 +106,10 @@ class FrameManager {
 
   private _handlePostMessage(handler: MessageHandler, event: MessageEvent) {
     const validated = validateIncoming(event.data);
+    const expectedClientOrigin = this._expectedClientOrigin();
     if (
-      event.origin === this._expectedClientOrigin() &&
+      expectedClientOrigin &&
+      event.origin === expectedClientOrigin &&
       event.source === this._iframe.contentWindow &&
       validated
     ) {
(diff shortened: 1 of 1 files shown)


src/FrameManager.ts
```diff
@@ -122,18 +124,23 @@ class FrameManager {
...
     try {
       const clientUrl = new URL(
         this._frameLocation,
         this._window.location.href
       );
-      return clientUrl.origin;
+
+      // Bail early if it's not an http resource.
+      // Ex: file protocol url origin behavior is browser specific
+      if (clientUrl.protocol !== 'http:' && clientUrl.protocol !== 'https:') {
```
justinaray: chrome actually blocked me from loading a local resource via a client config or js, but I figured better safe than sorry here.

src/FrameManager.ts
```diff
@@ -122,18 +124,23 @@ class FrameManager {
...
+        return null;
+      }
+
+      // Assert that this origin is actually valid.
+      // Ex: about:blank, chrome:version, extensions, etc will return 'null'
+      return new URL(clientUrl.origin).origin;
```
justinaray: Looks a little weird as linting didn't like a method invocation without an assignment.  "But javascript has side effects", I yelled.

Anyway, this handles the 'null' case of about:blank and others a little more generically.

src/FrameManager.ts
```diff
@@ -122,18 +124,23 @@ class FrameManager {
...
     } catch (err) {
-      return 'about:blank';
+      return null;
```
justinaray: Since `about:blank` isn't considered a valid origin by URL, I figured it made more sense to return null here.  This ensures we don't accidentally somehow match.  I also added truthy guards after this method is used.

PR #52, opened 2019-04-08
[COMUI-4] Generate documentation.

MattCheely:
There are also some small API changes to simplify the interface and/or clarify the docs.


MattCheely:
I got this branch deployed before PKO and then with everything going on before/during the conference forgot I never merged it!

.npmrc
```diff
@@ -0,0 +1,2 @@
+registry=https://registry.npmjs.org/
+@purecloud:registry=https://purecloud.jfrog.io/purecloud/api/npm/inin-internal-npm-dev/
```
stormojm: Did we want to expose this info into the public github?
Same with the jenkinsfile info

Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+        }
+      }
+    }
+
+    stage('Upload Docs') {
+      steps {
+        dir (env.REPO_DIR) {
+          sh '''
+             ./node_modules/.bin/upload \
+               --ecosystem gmsc \
```
stormojm: is this still the right ecosystem?

src/client.ts
```diff
@@ -17,31 +17,27 @@ import { Publication } from './messages/Publication';
...
-interface ClientConfigOptions {
-  clientWindow?: Window;
-}
-
-/**
- * The Client is access point for the embedded UI's in the host application.
- */
-class Client {
+export class Client {
   private _isStarted: boolean;
```
DKirwan: This is a nit pick: we typically avoid the `_` prefix in typescript because we can mark variables and functions as `private` explicitly

.npmrc
MattCheely: It's not ideal, but it is safe since Neither file has any actual credentials in it. I added the .npmrc to help stop package-lock.json from continuously getting the PureCloud URLs in it (which I do constantly, since that's the default for my personal .npmrc, along with most folks in PureCloud)

I'm open to other suggestions, but I'd prefer not to keep build/doc stuff in a separate repo, since I suspect that will just mean it gets out of date quickly. 

I could maybe improve things  bit for potential external devs by making the deploy tools an optional dependency?

Jenkinsfile
MattCheely: It is for now. I will eventually migrate the docs to PureCloud, but I'm saving that for a separate PR.

Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+      }
+    }
+
+    stage('Build Docs') {
+      steps {
+        dir (env.REPO_DIR) {
+          sh "npm run doc"
+          sh "./scripts/generate-deploy-files"
+          sh '''
+              export CDN_ROOT=$(./node_modules/.bin/cdn --ecosystem gmsc --manifest doc/manifest.json)
```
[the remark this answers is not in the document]
MattCheely: Eventually. I'll move all the docs for all of the projects at some point, but I want to get a working doc build into master before I start migrating.

src/client.ts
MattCheely: I wouldn't do this in a standalone project, but the expectation is that this library may be used from within JavaScript projects, at which point all the privacy annotations are gone, so we still need some obvious indicator that folks have stepped outside of supported behavior. Unfortunately, `_` prefixing is the best we have there without breaking a bunch of TypeScript conventions.

scripts/generate-deploy-files
```diff
@@ -0,0 +1,90 @@
...
+const deployVersion = require('../package.json').version;
+const manifestPath = path.join(docPath, 'manifest.json');
+const versionPath = path.join(docPath, 'versions.json');
+const versionsUrl =
+  'https://app.gmscdev.com/common-ui-docs/iframe-coordinator/versions.json';
+
+main();
+
+function main() {
+    let htmlFiles = glob.sync('**/*.html', { cwd: docPath })
```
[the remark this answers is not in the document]
MattCheely: Thanks for the catch. It looks like the prettier auto-format on save in spacemacs is doing some odd things, or failing to do them.

src/client.ts
DKirwan: Gotcha. Thanks!

PR #53, opened 2019-04-09
[COMUI-53] Limit origins the client will publish messages to.

MattCheely:
Consumers can provide a `hostOrigin` that is a valid embedding for the
client or when no value is provided, the client will only post messages
to a host on the same origin.


src/client.ts
```diff
@@ -16,6 +16,11 @@ import {
 import { Publication } from './messages/Publication';
 import { Toast } from './messages/Toast';
 
+interface ClientConfigOptions {
```
stormojm: TSDoc doesn't require the interface to be commented?

MattCheely: Actually, yes and there's a bunch of stuff tslint is complaining about related to properties not being documented.  Guess I've got some work to do...

PR #57, opened 2019-06-10
Add cli app

MattCheely:
This cleans up a few TS build issues, and adds a new cli tool intended to make it easier to develop iframe client applications. The CLI spins up a very basic Vue app with a `frame-router` element and injects some user-specified configuration.

Currently the host app doesn't have support for anything like notifications or arbitrary pub-sub. I will add those features in a future update (and probably eventually replace the demo app with it), but this will help folks get started while those additions are in flight.


cli/embedded-app/src/views/IframeEmbed.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <div id="routerLayout">
+    <div class="explainer">
+      Showing
+      <span class="app-route">{{ frameRoute }}</span>
```
MattCheely: I am considering having the `frame-router` emit a custom event like `transitioned` that indidcates what URL it's displaying internally. That information could be added here as a helpful sanity check. Something like:

`Showing /foo/bar as https://localhost:8080/clientapp/#/foo/bar`

cli/embedded-app/package-lock.json
```diff
@@ -0,0 +1,11345 @@
+{
+  "name": "wrapper-app",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://purecloud.jfrog.io/purecloud/api/npm/inin-internal-npm/@babel/code-frame/-/code-frame-7.0.0.tgz",
```
stormojm: I can never remember if we decided to leave the internal npm links or not...

cli/embedded-app/src/main.ts
```diff
@@ -0,0 +1,16 @@
+import Vue from 'vue';
+import App from './App.vue';
+import router from './router';
+import store from './store';
+
+import { registerCustomElements } from '../../../dist/index.js';
```
stormojm: It would be nice to have this act like documentation and be imported from '@purecloud/iframe-coordinator' (or whatever the package name is).  But I am not certain a way to do that without adding complexity of lerna or npm links...

cli/embedded-app/src/views/IframeEmbed.vue
stormojm: I think we have to add some way to test/show any events.  if you're testing your embedded app, you're going to want to see what happens when you send it messages or receive messages.

I'd almost envision a storybook style sidebar.

MattCheely: I agree completely. That's part of the feature set I want to add soon. At this point though, somebody can at least click around their app and manually verify it's working, and I wanted to go ahead and ship that.

cli/embedded-app/package-lock.json
MattCheely: Ugh, we decided not to, and I added an .npmrc file to try to avoid it, but of course it doesn't impact the nested project. I'll fix that. Good catch.

cli/embedded-app/src/main.ts
MattCheely: I tried that initially, but ran into some issues with babel trying to transpile the already transpiled module. I do have a better understanding of what was going on there now though, so maybe I can get the previous version working again. I'll give it a shot.

PR #58, opened 2019-06-19
Add a number of needed features to the iframe-coordinator host CLI.

MattCheely:
- Display of current route and frame location
- Toast notification support
- Client PubSub message display
- Better navigation support

This also moots the need for a separate host-app demo, so
that code was removed.


client-app-example/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+
+let babelRule = {
+  test: /\.js$/,
+  // Note: node_modules should not be excluded here to transpile dependencies
+  use: {
+    loader: 'babel-loader',
+    options: {
+      presets: [
+        [
+          '@babel/preset-env',
```
stormojm: do we want to give this a browser list?

client-app-example/src/client.js
```diff
@@ -0,0 +1,61 @@
...
+  document.getElementById('path').innerHTML = window.location.hash;
+};
+
+// Start intercepting link click events for routing
+let iframeClient = new Client({
+  // This lets the client work correctly with the cli host
+  hostOrigin: 'http://localhost:3000'
+});
+iframeClient.addListener('environmentalData', envData => {
+  document.getElementById('env-data').innerHTML = JSON.stringify(
```
stormojm: It might be cool, to show us setting the Intl API to the locale or something to show some real examples of what needs to be done.

MattCheely: That's a pretty good idea. I'll add something like that.

cli/example-ifc.config.js
```diff
@@ -12,8 +16,12 @@ module.exports = function(frameRouter) {
       custom: getCustomClientData()
     }
   );
+
+  frameRouter.messaging.addListener('publish.topic', publication => {
+    hostActions.notifyPubSub(publication);
+  });
```
MattCheely: I set up this `notifyPubSub` hook because the host app can't know in advance what topics the client will send, but after thinking about it this morning, I think rather than providing a hook to show pubSub events, it would be better for the setup function to just return some config like: 
```
{
   clientTopics: [ 'publish.topic' ]
}
```

PR #61, opened 2019-06-26
Registered key publications

JeffRiggle:
This is an initial stab at sending key events in some format from the iframe to the host. In this approach you request the keys represented as strings that you would like updates for and the frame will expose those as they are encountered.

The big drawbacks of this approach are. keys represented as strings especially complex ones are hard. For example 'aalt' or 'aaltshift'. Also being specific about what you want might be annoying since `a`, `a+alt`, `a+ctrl`, `a+shift`, `a+shift+alt` and `a+shift+ctrl` are all different keys.

I have another approach that using a form of filtering to send events so I can put that branch up for draft PR as well if we would like to review that approach.


src/client.ts
```diff
@@ -111,6 +112,28 @@ export class Client {
...
+      (key: string) => key === keyData.serialize()
+    );
+    if (!shouldSend) {
+      return;
+    }
+
+    this._sendToHost({
+      msgType: 'publish',
+      msg: {
+        topic: 'keydown.topic',
```
JeffRiggle: This might be an abuse of the topic system so I can change to do something else

src/Key.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  return event.key;
+}
+
+/**
+ * Key class for converting and handling key events.
+ */
+class Key {
```
JeffRiggle: Honestly this could probably be a module itself but I do not think we want to break this out and spread that dependency around.

cli/example-ifc.config.js
```diff
@@ -13,14 +13,15 @@ module.exports = function(frameRouter) {
     {
       locale: 'en-US',
       hostRootUrl: window.location.origin,
+      registeredKeys: ['a', 'b', 'actrl', 'ashift', 'actrlshift'],
```
covertcj: I'm not a huge fan of this key format.  Any reason we can't do something like `ctrl+shift+a`?  That seems much more intuitive/more consistent with other programs that do key bindings.

Looking forward to other parts of the review, it seems like if I registered `aaltctrl` that would just silently fail, right?  We might want to do some sort of parsing/verification on these keys.

justinaray:
Keyboard events have been notoriously difficult cross-platform/devcie and have had a tumultuous history.  I think we're getting closer to platform-agnostic events with the latest spec so I would definitely recommend modeling our api after [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent).  For instance, instead of trying to stringify them, just let the registration be object-based.

Is your filtering branch more closely aligned to the spec?

cli/example-ifc.config.js
JeffRiggle: Yeah I could change the format and add validation. The other part that is not great currently is ordering matters. `aaltctrl` is not `ctrlalta`. Really I do not like having this list of string in general and maybe the solution is using the Key class as the list but I am not sure how that would work for our consumers.

justinaray:
Also, wondering about the use-cases that client-to-host and host-to-client would provide.  I could see host-to-client for hotkeys, but would be interested in hearing all the cases here.

i.e. wondering if the juice is ultimately worth the squeeze.  Very well might be, but worth discussing if we should.

JeffRiggle:
This Key class an attempt to handle all sorts of permutations of the KeyEvent. As you stated the key event is difficult to say the least and has options that are deprecating (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) has tons of info.

Basically the string is just a serialized (easy to compare) format of a more complex object. I think the Key object should be a relatively close match to KeyboardEvent.

As far as filtering code goes. In that case I took an approach that uses complex objects to do basically a sort of reflection comparison.

With that approach you would tell it what topic you are filtering (because topic was not granular enough) and how you are filtering it for example

`
// match anything that does not have the altKey
filteredTopics: {
          'keydown.topic': {
            filters: [
              {
                property: 'altKey',
                comparison: 0,
                expected: 'false'
              }
            ],
            junction: 'and'
          }
        }
`

I can put that up for PR as well since that is its own branch in my fork

JeffRiggle:
In my opinion when you want to start sending key data from the host to the client you are moving into the I own the full process and I know exactly how I plan to interact with my child territory.

So I would think that you could use the existing API's to register custom message passing from the host to the client.

However I could be wrong on this

cli/example-ifc.config.js
covertcj: Yeah, that was part of the reason I thought something like `+` delimited lists made more sense.  It seems like it would be fairly easy `.split('+')` it and do simple validation/order independent parsing

src/client.ts
stormojm: I think i would prefer a direct message instead of the generic publish, since this is an Iframe event and not an app event.  It seems similar to  requestToast in that regards.

cli/example-ifc.config.js
stormojm: you would have to worry about using + as your key with that approach :)

But i agree this format should change.

cli/embedded-app/src/App.vue
```diff
@@ -2,6 +2,7 @@
   <div id="app">
     <notifications group="toast" id="toastNotifications" position="top right"></notifications>
     <notifications group="pubsub" id="pubsubEvents" position="top left"></notifications>
+    <notifications group="keydown" id="keydowmEvents" position="bottom left"></notifications>
```
stormojm: typo on the id

src/messages/Lifecycle.ts
```diff
@@ -53,6 +55,7 @@ const envDataDecoder = guard(
   object({
     locale: string,
     hostRootUrl: string,
+    registeredKeys: mixed,
```
stormojm: you should be able to give this a stronger guard.  Its just an array of strings, right?

cli/example-ifc.config.js
```diff
@@ -13,6 +13,13 @@ module.exports = function(frameRouter) {
     {
       locale: 'en-US',
       hostRootUrl: window.location.origin,
+      registeredKeys: [
+        'a',
+        'b',
+        'actrl',
+        'ashift', 
+        'actrlshift'
```
MattCheely: I believe there was some previous conversation about this, but I think a better format here is to just mimic the structure of a few non-controversial parts of the KeyboardEvent spec.  It will mean less logic to decode strings into a logical representation, and won't require folks to remember another way of representing key presses outside of the browser's API.

Something like:
```
{ 
  key: 'a', 
  shiftKey: false, 
  altKey: false, 
  ctrlKey: false, 
  metaKey: false 
}
```
or just `a` or `A` for short versions with no modifers or only shift (if we really need to support that).

MattCheely:
Do we think we need to support access to bare key events without modifier keys?  There's an argument to be made against it in that in a cross-domain context, it can turn the host app into a keylogger. It's not a security issue for our use cases, but if we leave it out it makes the library as a whole useful in more contexts.

cli/example-ifc.config.js
JeffRiggle: Yeah I am fine with the above structure and I even started work on that but I wanted to wait for more feedback. The reason why I am including the modifier keys very explicitly is because of interesting cross environment details. For example in Mac `alt+f` is ƒ but in Windows its just f which makes cross environmental details a nightmare if you are not overly explicit.

src/Key.ts
```diff
@@ -0,0 +1,147 @@
...
+  /** The key (character) that was pressed */
+  public key: string;
+  /** If the alt key was pressed */
+  public alt?: boolean;
+  /** If the ctrl key was pressed */
+  public ctrl?: boolean;
+  /** If the shift key was pressed */
+  public shift?: boolean;
```
MattCheely: In this case as well, I'd just use a subset of the data structure provided by `KeyboardEvent`, and following the same rules.
```
{ 
  key: string,
  shiftKey: boolean, 
  altKey: boolean, 
  ctrlKey: boolean, 
  metaKey: boolean 
}
```
No need to mentally switch between property names, very little extra to remember outside of the browser API.

JeffRiggle:
> Do we think we need to support access to bare key events without modifier keys? There's an argument to be made against it in that in a cross-domain context, it can turn the host app into a keylogger. It's not a security issue for our use cases, but if we leave it out it makes the library as a whole useful in more contexts.

I am not 100% sure I follow the request. Right now we will send the barebones data about the key pressed. We do not send any other information other than (key, code, keyCode, alt, ctrl and shift). There are other bits of information attached to the key event like element; However, those cannot be posted to the hosted application because the browser locks it down as a CORS exception. If there are other specific details you think should be sent along just let me know

MattCheely:
Sorry

> > Do we think we need to support access to bare key events without modifier keys? There's an argument to be made against it in that in a cross-domain context, it can turn the host app into a keylogger. It's not a security issue for our use cases, but if we leave it out it makes the library as a whole useful in more contexts.
> 
> I am not 100% sure I follow the request. 

Sorry, I'll try to clarify. Right now, the host could ask for key events to be sent for any unmodified or shift-only keypresses of a-zA-Z0-9{punctuation} and log everything the user types into every input field in the client app, acting as as a keylogger that could capture sensitive data even if the client does not trust the host.  However, if we only allow key events to be transmitted from client to host if a ctrl, alt, or meta modifer key is pressed, then there's no way to read data like that.  It seems to me that the primary use case of keyboard shortcuts would still work under those conditions.

cli/example-ifc.config.js
MattCheely: That's fair, but since the host app presumably also has to handle those platform variations for native key events, wouldn't it be simpler to just have the client key events mimic the native version and let the host app use the same logic for both native events and iframe events? My goal is generally to give folks as uniform of an API as possible, even when that means mimicking an API that I don't like.  I'd rather have one interface that's just ok than two or more of varying quality.  

I would reconsider that though if things like angular/vue/etc somehow normalize browser key events behind the scenes and they are doing it consistently.

JeffRiggle: So there are two parts to this and I think an exact mimic is going to be hard to pull off over time especially as the spec changes. The two parts are the registration and the propagation. 

In the case of the registration admittedly it is a bit easier and less error prone to limit the options to the above. The options we are leaving out are code (not yet supported in IE and Edge) and keyCode (supposedly deprecated for code). The question becomes should we be helping developers map intents by hiding some complexity? Also what do you do when the developer provides code, keyCode and key what is your actual comparison at that point? In my opinion this specific part should probably stay as the above object.

In the case of propagation back to the host I will admit that maybe we could be a bit more consistent and provide the un-translated key as well as the properties code and keyCode. I think the benefit that angular and vue have over our use is they can just provide the raw keyboard events where we cannot because we are going cross origin. The only concern I have with making this very specific to browser detail is every time the keyboard event spec changes we will have to update our implementation to send more data.

cli/embedded-app/src/views/IframeEmbed.vue
```diff
@@ -10,6 +10,7 @@
       id="frameRouter"
       v-bind:route="frameRoute"
       v-on:toastRequest="displayToast"
+      v-on:keyDown="handleKeyEvent"
```
stormojm: keydown as an event name seems a little off to me since it could be confused with normal key down.  I think it might be nice to tie it to the registeredKeys you used to subscribe to them.

src/messages/Lifecycle.ts
```diff
@@ -53,6 +71,7 @@ const envDataDecoder = guard(
   object({
     locale: string,
     hostRootUrl: string,
+    registeredKeys: array(mixed),
```
stormojm: you should be able to guard this array better

cli/embedded-app/src/views/IframeEmbed.vue
JeffRiggle: As in the name should have some mention of registeredKeys or something like that? For example registeredKeyFired

cli/example-ifc.config.js
MattCheely: To be clear, I'm not suggesting that we pass through or use any data besides the interface:
```
{ 
  key: string,
  shiftKey: boolean, 
  altKey: boolean, 
  ctrlKey: boolean, 
  metaKey: boolean 
}
```

keyCode, code, etc. are unnecessary for this interface at the point, and we should leave them out unless we have a specific use for them. There's no need to worry about those fields changing in the browser spec anytime in the foreseeable future.

Reading between the lines a bit, I think your primary concern is that to handle `alt+f` across OS, developers would have to register two keys to handle the case where OSX returns ƒ. Is that correct?  If so, I think it's worth noting that using the option(alt) key for OSX keyboard shortcuts breaks the OS conventions and that command is probably more appropriate in that case. As I understand it, the point of the option key is to provide a way to input alternate characters, so masking that seems strange.  Worst case, it's also not the end of the world to have to do some extra specification. 

I'm fine if we want to make the match against `key` case-insensitive, but I wouldn't go quite so far as changing the character involved.

src/Key.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /** Tests equalality of another key. */
+  public equals(key: Key): boolean {
+    return key.serialize() === this.serialize();
+  }
+}
+
+/** Data structure representing a native key event. */
+interface NativeKey {
```
MattCheely: Do we really need to send `charCode`, `code` and `keyCode`? What are they used for?

src/client.ts
```diff
@@ -111,18 +114,64 @@ export class Client {
...
   };
 
+  private _onKeyDown = (event: KeyboardEvent) => {
+    if (!this._registeredKeys) {
+      return;
+    }
+
+    const keyData = Key.fromKeyEvent(event);
+    const shouldSend = this._registeredKeys.some(
+      (key: string) => key === keyData.serialize()
```
MattCheely: Serialization seems like an unnecessary extra step for comparison of the events. It's just more code to ship to the browser and extra logic where bugs could creep in.  Seems like just comparing the individual fields from the key registration data should be sufficient.

JeffRiggle: Yeah the thought was that string comparison should be less performance intensive than multiple property comparison since we could serialize the registered keys at startup. But yeah it is more code and maybe reducing the load on v8 would result in faster speeds. I went ahead and changed this

cli/example-ifc.config.js
JeffRiggle: I could be missing something here but I think code and keyCode are pretty important. I agree that most of the cases you could blame alt as the mapping key but I think that is a valid use case. 

The following are some examples of confusing behavior with just those properties `alt+f`, `alt+n` and `alt+e`. The big issue is with `alt+n` and `alt+e`. In the case of windows you would get key = n/e alt = true. In the case of mac you would get key = dead alt = true for both. The only real way to distinguish these two events on mac is using code (KeyN/KeyE) or keyCode (732/180)

The way I have it now seems like it should handle all cases pretty well. You register the key using its raw value (e/f/n) and any modifiers; However, when you get the event back from the API it provides key in its raw form along with code and keyCode to help you decide how to handle that event.

src/Key.ts
MattCheely: I ask because if we don't actually need to send those, the `Key` class is largely unnecessary. At that point it's just renaming fields and handling a comparison, which could easily just be a helper function.

JeffRiggle: I think I mentioned this in a different section of this review. I will copy it over here. Let me know if you have any additional feedback on this.

> I could be missing something here but I think code and keyCode are pretty important. I agree that most of the cases you could blame alt as the mapping key but I think that is a valid use case.
> 
> The following are some examples of confusing behavior with just those properties `alt+f`, `alt+n` and `alt+e`. The big issue is with `alt+n` and `alt+e`. In the case of windows you would get key = n/e alt = true. In the case of mac you would get key = dead alt = true for both. The only real way to distinguish these two events on mac is using code (KeyN/KeyE) or keyCode (732/180)
> 
> The way I have it now seems like it should handle all cases pretty well. You register the key using its raw value (e/f/n) and any modifiers; However, when you get the event back from the API it provides key in its raw form along with code and keyCode to help you decide how to handle that event.

JeffRiggle: Misread this. I see so you would like to get rid of the class. Yeah I could look into that.

JeffRiggle: Key class is now gone in favor of a helper function. Let me know if there are any other requested changes or feedback

PR #62, opened 2019-06-26
documentation updates

JeffRiggle:
Just wanted to give some feedback on the README.md. These are all things I kind of stumbled through on the first little bit of working with this. The biggest issue I had was my node version was 12.4.x. I figure this would just be generally useful

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,40 @@ function getCustomClientData() {
 }
 ```
 
+### Building
+
+#### Installation
+Before you can build this you will need to make sure that you are using a node version that matches our target version of node. To see this version you can check the [Jenkinsfile](Jenkinsfile)
+
+##### On Mac
+It would be recommended to manage node version using something like nvm
+```
+brew install nvm
+nvm use 10.15.3
+npm ci
+```
+
+##### On Windows or Linux
+Make sure the correct version of node is being used and run install
+```
+node --version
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -192,6 +192,40 @@ function getCustomClientData() {
...
+##### On Mac
+It would be recommended to manage node version using something like nvm
+```
+brew install nvm
+nvm use 10.15.3
+npm ci
+```
+
+##### On Windows or Linux
+Make sure the correct version of node is being used and run install
```
covertcj: What's the 'correct' version of node?  Was 12.4 just too old?

covertcj: Oh, nevermind, I can see that above... Why is 10.15.3 necessary?

JeffRiggle: 10.15.3. 12.4 was too new. I think this had something to do with our dependencies something like a compile of node-gyp failed. I imagine we will eventually want to look into that as well

MattCheely: I would just note that consumers should build with the latest node LTS version and leave it at that. We should be making sure it works consistently as new releases come out, and we won't need to update the docs every time there's a new release.

I'd also leave out specific brew/nvm steps for setting up a node version. We can tell folks what version they should use, but there's no real reason for us to maintain instruction on how to manage node.

PR #65, opened 2019-07-09
Add an ES5 build and fix issues with examples in IE11.

MattCheely:
We hit an issue in the PureCloud UI where the iframe-coordinator library wasn't IE11 compatible due to a `class` keyword. Because it's often a pain to get pre-packaged node dependencies transpiled to ES5, the best solution seemed to me to be to ship an additional es5 version in the library.

### DIFF
--- a/cli/embedded-app/babel.config.js
+++ b/cli/embedded-app/babel.config.js
@@ -1,4 +1,13 @@
 module.exports = {
-  presets: ['@vue/app'],
-  exclude: ['../../dist/*.js']
+  presets: [
+    '@vue/app',
+    [
+      '@babel/preset-env',
+      {
+        useBuiltIns: 'entry',
+        targets: { browsers: 'ie 11, defaults' }
+      }
+    ]
+  ],
+  exclude: ['../../dist/**/*.js']
 };
--- a/cli/embedded-app/src/main.ts
+++ b/cli/embedded-app/src/main.ts
@@ -1,10 +1,16 @@
+// Required polyfills for iframe-coordinator IE11 support
+import 'core-js/stable';
+import 'custom-event-polyfill/polyfill';
+import '@webcomponents/custom-elements/src/native-shim';
+import '@webcomponents/custom-elements';
+// IE11 compatible iframe-coordinator import
+import { registerCustomElements } from 'iframe-coordinator/dist/es5';
+
 import Vue from 'vue';
 import Notifications from 'vue-notification';
 import App from './App.vue';
 import router from './router';
 import store from './store';
-
-import { registerCustomElements } from 'iframe-coordinator';
 
 registerCustomElements();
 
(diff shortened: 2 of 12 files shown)


cli/embedded-app/babel.config.js
```diff
@@ -1,4 +1,15 @@
...
-  presets: ['@vue/app'],
-  exclude: ['../../dist/*.js']
+  presets: [
+    '@vue/app',
+    [
+      '@babel/preset-env',
+      {
+        useBuiltIns: 'entry',
+        targets: {
+          ie: '11'
```
stormojm: Its probably not a huge deal.  But this could in theory remove conversions that firefox/chrome need but IE doesn't. 
 It's more accurate to include all of your targets.

cli/embedded-app/src/main.ts
```diff
@@ -1,11 +1,17 @@
+// Required polyfills for iframe-coordinator IE11 support
+import 'core-js/stable';
+import 'custom-event-polyfill/polyfill';
+import '@webcomponents/custom-elements/src/native-shim';
+import '@webcomponents/custom-elements';
+// IE11 compatible iframe-coordinator import
```
stormojm: Could we add a top shim layer that auto picked the right dist?  I guess that would make it a bit bloaty for apps that wouldn't need the es5 version...

package-lock.json
```diff
@@ -513,7 +512,7 @@
     },
     "array-flatten": {
       "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/array-flatten/-/array-flatten-1.1.1.tgz",
+      "resolved": "https://purecloud.jfrog.io/purecloud/api/npm/inin-internal-npm/array-flatten/-/array-flatten-1.1.1.tgz",
```
stormojm: switced to purecloud

tsconfig.es5.json
```diff
@@ -0,0 +1,13 @@
+{
```
stormojm: you could commonize some of these properties by extending the base tsconfig.

webpack.es5.config.js
```diff
@@ -0,0 +1,7 @@
+const path = require('path');
+const config = require('./webpack.config.js');
+
```
stormojm: Consider using webpack-merge to make this cleaner looking.

cli/embedded-app/babel.config.js
stormojm: maybe just use the new defaults query 7.5 added? https://babeljs.io/blog/2019/07/03/7.5.0#defaults-browserlist-query-in-babel-preset-env-8897-https-githubcom-babel-babel-pull-8897

MattCheely: Ah yeah, that was me rushing / being lazy. Thanks for catching me on it.

package-lock.json
MattCheely: facepalm.jpg

cli/embedded-app/src/main.ts
MattCheely: Yeah, we'd have to include both version of the lib in the app build, which would be overkill, IMO.

webpack.es5.config.js
MattCheely: Of course webpack configuration is so complicated somebody had to write a merge tool for extending it. Sigh. I'm gong to let my grump flag fly on this one and say I'd rather not have the additional complexity of more hidden code affecting the build, but it could be worth revisiting if the base configuration gets more complicated in the future.

PR #66, opened 2019-07-16
Pass assigned route to client via env data

zsdsjl:
This PR allows client to know its own assigned route in the host app. Then the client can know its real URL in user's browser with `hostRootUrl` and `assignedRoute`.

With this PR, in sample app, the client side can see the `assignedRoute `.

image

Do we need to pass the whole url instead of just assigned route? because the real URL in browser would contain # in the url.

If the host always uses the hash router, then only `assignedRoute` and `hostRootUrl` should be enough for client.

### DIFF
--- a/src/HostRouter.ts
+++ b/src/HostRouter.ts
@@ -22,14 +22,16 @@ export class HostRouter {
   public getClientTarget(route: string): ClientTarget {
     let clientTarget: ClientTarget = {
       id: null,
-      url: null
+      url: null,
+      assignedRoute: null
     };
     this._clients.forEach(client => {
       const clientRoute = matchAndStripPrefix(route, client.assignedRoute);
       if (clientRoute !== null) {
         clientTarget = {
           id: client.id,
-          url: applyRoute(client.url, clientRoute)
+          url: applyRoute(client.url, clientRoute),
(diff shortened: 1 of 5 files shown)


stormojm:
What's the use case for this change?  The routes could be different per host application, so the client really shouldn't change its logic based on the route.

MattCheely:
The main thing we want to do is enable clients to generate link URLs that match the top-level route used by the host app.  That said, @zsdsjl, I think it would be nice if we don't expose the client route directly to the app, but instead have the client library track it privately and use it to provide a function for translating urls/paths.

zsdsjl:
@stormojm As @MattCheely mentioned, we will want to render some link in client app allowing user to bookmark or share the URL in the browser.

@MattCheely I think in order to do like this, we have two ways:
1.  Provide an async function: an async function on the client side will send a request to host and host response the assignedRoute. Then this function and do the translation and resolve the translated one.
something like `client.getBrowserUrl().then(url => {})`

2.  provide a simple sync function: Client library can get `assignedRoute` silently and store it as private variable. And then a simple api function can read it.
like `const url = client.getBrowserUrl(); // read from the client instance private member variable`

I think the second way is what you are saying, right? @MattCheely

PR #68, opened 2019-07-18
Add route assenged as private

zsdsjl:
Created a new branch for the same purpose as https://github.com/purecloudlabs/iframe-coordinator/pull/66

This time the host will still pass the `assignedRoute` data in the `"env_init"` response.

And the client will extract the `assignedRoute` from the response and store it as a private variable.
Then expose the rest env data.

And client and use `client.hostURL` getter to get the whole current host url.

Is this `client.hostURL` a good getter name?

### DIFF
--- a/client-app-example/ifc-cli.config.js
+++ b/client-app-example/ifc-cli.config.js
@@ -14,10 +14,10 @@ module.exports = function(frameRouter) {
     },
     {
       locale: 'en-US',
-      hostRootUrl: window.location.origin,
+      hostRootUrl: window.location.origin + '/#/',
       registeredKeys: [
         { key: 'a', ctrlKey: true },
-        { key: 'b', altKey: true }, 
+        { key: 'b', altKey: true },
         { key: 'a', ctrlKey: true, shiftKey: true }
       ],
       custom: getCustomClientData()
--- a/client-app-example/public/client-app-1/index.html
+++ b/client-app-example/public/client-app-1/index.html
@@ -13,10 +13,12 @@
 
     <h3>Change Route</h3>
     <ul>
-      <li><a href="#/app1/route1">This client: /route1</a></li>
-      <li><a href="#/app1/route2">This client: /route2</a></li>
-      <li><a href="#/app2/route1">Other client: /route1</a></li>
-      <li><a href="#/app2/route2">Other client: /route2</a></li>
+      <!-- internal routes URLs get remapped in js by `transformLinks` -->
+      <li><a href="#/route1">This client: /route1</a></li>
+      <li><a href="#/route2">This client: /route2</a></li>
+      <!-- External URLs use full links so they open nicely in new tabs -->
+      <li><a href="localhost:3000/#/app2/route1">Other client: /route1</a></li>
+      <li><a href="localhost:3000/#/app2/route2">Other client: /route2</a></li>
     </ul>
 
     <h2>Toasts</h2>
--- a/client-app-example/public/client-app-1/index.html
+++ b/client-app-example/public/client-app-1/index.html
@@ -8,6 +8,8 @@
     <h2>Routing</h2>
     <h3>Current Local Route:</h3>
     <div id="path"></div>
+    <h3>Current Host Route:</h3>
+    <div id="hostPath"></div>
 
     <h3>Change Route</h3>
     <ul>
--- a/client-app-example/src/client.js
+++ b/client-app-example/src/client.js
@@ -9,6 +9,7 @@ document.getElementById('path').innerHTML = window.location.hash;
 
 window.onhashchange = function() {
   document.getElementById('path').innerHTML = window.location.hash;
+  document.getElementById('hostPath').innerHTML = iframeClient.asHostUrl(window.location.hash);
 };
 
 /****  SET UP THE IFRAME CLIENT LIBRARY ****/
@@ -30,7 +31,7 @@ iframeClient.addListener('environmentalData', envData => {
   console.log(
     `Got locale from host. Current date formatted for ${envData.locale} is: ${localizedDate}`
   );
-  console.log(`current host url is ${iframeClient.hostURL}`)
+  document.getElementById('hostPath').innerHTML = iframeClient.asHostUrl(window.location.hash);
   displayEnvData(envData);
 });
 
--- a/client-app-example/src/client.js
+++ b/client-app-example/src/client.js
@@ -9,7 +9,9 @@ document.getElementById('path').innerHTML = window.location.hash;
 
 window.onhashchange = function() {
   document.getElementById('path').innerHTML = window.location.hash;
-  document.getElementById('hostPath').innerHTML = iframeClient.asHostUrl(window.location.hash);
+  document.getElementById('hostPath').innerHTML = iframeClient.asHostUrl(
+    window.location.hash
+  );
 };
 
 /****  SET UP THE IFRAME CLIENT LIBRARY ****/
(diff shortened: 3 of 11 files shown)


src/client.ts
```diff
@@ -56,6 +57,7 @@ export class Client {
     );
     this._envDataEmitter = new InternalEventEmitter<EnvData>();
     this._registeredKeys = [];
+    this._assignedRoute = '';
```
MattCheely: Typically, I'd opt for a type of `string | null` and initialize with `null` instead of an empty string.  It keeps the representation of "no data yet" uniform across types.

src/client.ts
```diff
@@ -185,6 +189,13 @@ export class Client {
     return this._environmentData;
   }
 
+  /**
+   * Gets the host url prefix for current app
+   */
+  public get hostURL() {
+    return `${this.environmentData.hostRootUrl}/#/${this._assignedRoute}`;
```
MattCheely: I think we probably shouldn't assume the host app is using hash-based routing.  It's probably better if we ask the host to incorporate the `/#/` part into the data it sends for `hostRootUrl`.

Rather than a simple getter here, I was thinking of something more like:
```
asHostUrl(clientRoute) {
  return path.join(this.environmentData.hostRootUrl, this._assignedRoute, clientRoute);
}
```
obviously, we don't have `path.join` in the browser, but there are some related functions around handling leading/trailing slashes in `HostRouter.ts` that might be refactored into some common path utilities.

MattCheely:
It would be nice to see updates to the client-app-example so that the links in the demo point to the right host app URLs and behave nicely when opened in a new tab.

src/client.ts
zsdsjl: make sense to me

src/client.ts
zsdsjl: Agree, I don't like hardcoded '#' as well. I could try to make the "hostRootUrl" pass the whole url  before the `assignedRoute`

zsdsjl:
> It would be nice to see updates to the client-app-example so that the links in the demo point to the right host app URLs and behave nicely when opened in a new tab.

Will update demo app as well

zsdsjl:
In the latest changes:
1.  Put path helper function into a util module
2.  The `hostRootUrl` will contain `#` if the host app uses hash or the `hostRootUrl` in config file contains `#`. 
3.  provides client a `asHostUrl` method. Passing client route string, it will return the current host app url
4.  Shows the current host url in the demo app

src/client.ts
```diff
@@ -185,6 +190,18 @@ export class Client {
     return this._environmentData;
   }
 
+  /**
+   * Gets the host url prefix for current app
+   */
+  public asHostUrl(clientRoute: string) {
```
stormojm: minor nit: we have so far kept return types on all methods.  So :void in this case.

src/messages/Lifecycle.ts
```diff
@@ -46,6 +46,14 @@ export interface EnvData {
   custom?: any;
 }
 
+/**
+ * Extended Environmental data passed to client with some private data.
+ */
+export interface EnvDataExt extends EnvData {
```
stormojm: I am not a fan of names like this (e.g Extended or Ext).  It doesn't give any meaning to the actual class and leads to people just dumping things into it because it has a generic name/purpose.

src/client.ts
stormojm: Also can we add some tests to this

src/urlUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Removes any trailing '/' characters from a string.
+ *
+ * @external
+ */
+export function stripTrailingSlash(str: string): string {
+  return str.replace(/\/+$/, '');
+}
+
+/**
+ * Removes any leading '/' or '#' characters from a string.
```
stormojm: do we have to worry about query strings in urls?

src/client.ts
zsdsjl: Will add test cases

src/urlUtils.ts
zsdsjl: This function is for removing the leading / and # of the client route so that client can concat the host url and client route.  As query string is at the tail of the client route, we don't need to worry about it in this case.
And for this function itself, the purpose is just to remove all slash and hashtag. so it should be okay

src/messages/Lifecycle.ts
MattCheely: How about something like `BooststrapData` or `SetupData`?

zsdsjl: I am pretty bad at naming variables. The purpose of having two kind of `EnvData` is that `EnvData` is the data exposed to client app and `ExtEnvData` contains `EndData` and some other private properties. Maybe change `ExtEnvData` to `RawEnvData`?

zsdsjl: > How about something like `BooststrapData` or `SetupData`?

Sounds great!

client-app-example/ifc-cli.config.js
```diff
@@ -14,10 +14,10 @@ module.exports = function(frameRouter) {
     },
     {
       locale: 'en-US',
-      hostRootUrl: window.location.origin,
+      hostRootUrl: window.location.origin + '/#/',
```
MattCheely: The host app now needs to pass the hash part of the URL down to the client app, since it's an always-present part of the host routing, and a key part of generating host URLs

client-app-example/public/client-app-1/index.html
```diff
@@ -8,13 +8,17 @@ <h1>Component #1</h1>
...
-      <li><a href="#/app1/route1">This client: /route1</a></li>
-      <li><a href="#/app1/route2">This client: /route2</a></li>
-      <li><a href="#/app2/route1">Other client: /route1</a></li>
-      <li><a href="#/app2/route2">Other client: /route2</a></li>
+      <!-- internal routes URLs get remapped in js by `transformLinks` -->
+      <li><a href="#/route1">This client: /route1</a></li>
+      <li><a href="#/route2">This client: /route2</a></li>
+      <!-- External URLs use full links so they open nicely in new tabs -->
+      <li><a href="localhost:3000/#/app2/route1">Other client: /route1</a></li>
+      <li><a href="localhost:3000/#/app2/route2">Other client: /route2</a></li>
```
MattCheely: I updated these links to use `href` fields from the client app's perspective, and then transform as necessary in javascript so folks running the demo can see what the generated links should look like, and when to use code to generate URLs vs just pointing at a new location.

client-app-example/src/client.js
```diff
@@ -79,6 +89,11 @@ document.addEventListener('DOMContentLoaded', () => {
 });
 
 // HELPER FUNCTIONS
+function transformLinks() {
+  document.querySelectorAll('a[href^="#"]').forEach(link => {
+    link.href = iframeClient.asHostUrl(link.getAttribute('href'));
+  });
+}
```
MattCheely: This is the code that maps internal route links to top-level URls.

client-app-example/src/client.js
```diff
@@ -23,12 +26,19 @@ let iframeClient = new Client({
...
 
+  const appLocale = envData.locale;
   const now = new Date();
   const localizedDate = new Intl.DateTimeFormat(appLocale).format(now);
   console.log(
-    `Got locale from host. Current date formatted for ${envData.locale} is: ${localizedDate}`
+    `Got locale from host. Current date formatted for ${
+      envData.locale
+    } is: ${localizedDate}`
+  );
```
MattCheely: Looks like my editor's prettier plugin got ahold of this line and changed some formatting.

zsdsjl:
The latest changes:
1.  Variable name changed
2.  Unit test added
3.  return type added

PR #69, opened 2019-07-19
A humble request

JeffRiggle:
For our use case `position: relative` causes us issues and in order to override this styling in css we have to `!important` it. I would think that styling should be left completely up to the consumer of the frame-router.

### DIFF
--- a/cli/embedded-app/src/views/IframeEmbed.vue
+++ b/cli/embedded-app/src/views/IframeEmbed.vue
@@ -103,7 +103,6 @@ for more details.
   height: 100%;
 }
 #frameRouter {
-  position: relative;
   flex-grow: 1;
   flex-shrink: 1;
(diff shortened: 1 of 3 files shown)


MattCheely:
Setting the position to relative is part of what allows the actual `iframe` element inside to render at the full size of the `frame-router` element. I'm fine with changing the approach, but it needs to be to something that preserves the relationship between the size of `frame-router` and the `iframe` inside it.

JeffRiggle:
Uploaded new approach, maybe this will work better. The styling should be the same except now we use an inline style instead of setting style attributes on the element. This should allow consumers to override the style using a more specific style defined in their own style sheets.

MattCheely:
This looks pretty good to me. I'm going to do some local tests and I'll merge if everything looks good.

PR #70, opened 2019-08-01
Add nav request

MattCheely:
This adds an API clients can use to request navigation programmatically. It's useful for redirect-like behaviors or cases where the basic link interception provided by the client doesn't work with the client framework.

### DIFF
--- a/src/specs/FrameManager.spec.ts
+++ b/src/specs/FrameManager.spec.ts
@@ -13,7 +13,10 @@ describe('FrameManager', () => {
         .createSpy('docCreateElement')
         .and.callFake((element: string) => {
           return mocks.frame;
-        })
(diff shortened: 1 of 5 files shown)


src/specs/FrameManager.spec.ts
```diff
@@ -13,7 +13,10 @@ describe('FrameManager', () => {
         .createSpy('docCreateElement')
         .and.callFake((element: string) => {
           return mocks.frame;
-        })
+        }),
+      head: {
+        appendChild: jasmine.createSpy('headAppend')
+      }
```
MattCheely: This is a test fix related to a previous commit that changed how we handle styling the frame-router internals.

PR #73, opened 2019-10-15
Comui 165 better failure modes

MattCheely:
A few folks using the library noticed that without `registeredKeys` in the host app's environment data, things were silently failing. There are two issues here:

1. registeredKeys should probably be optional
2. silent failures are really hard to debug in production

I have updated the message validation API to throw exceptions when bad messages are passed from host to client or vice versa. The exceptions have pretty nice error messages that should make debugging problems easier. They look like:

```
Error: I received invalid data to send to the host application. This is probably due to
bad input into one of the iframe-coordinator client methods.

{
  "msgType": "publish",
  "msg": {
    "data": "I forgot the topic",
  },
         ^ Missing key: "topic"
}
```

I'm inclined to think this should probably be a major version bump, even though it's unlikely that the new exceptions cause problems for the client or host apps. Thoughts?


client-app-example/ifc-cli.config.js
```diff
@@ -15,11 +15,13 @@ module.exports = function(frameRouter) {
     {
       locale: 'en-US',
       hostRootUrl: window.location.origin + '/#/',
+      /*
```
stormojm: Did you want to keep this commented out?

MattCheely: I dd not! Good catch!

src/messages/Lifecycle.ts
```diff
@@ -41,7 +38,7 @@ export interface EnvData {
   /** Location of the host app */
   hostRootUrl: string;
   /** Keys to notify changes on */
-  registeredKeys: KeyData[];
+  registeredKeys?: KeyData[];
```
JackieChiles: :100:

PR #74, opened 2019-10-16
Change "toast" api to "notification" for clarity.

MattCheely:
"Toast" as a term is more about a UX implementation than an intent,
and may be confusing for non-native english speakers who might wonder
what lightly cooked bread has to do with webapps.

I'm targeting this at my other PR, since I think that should be a major version bump.  This is another change that I think is worth making, and a major version bump prior to production use is probably a good spot for it. 

I've made backwards compatibility affordances in the decoders, at least in the event that the host upgrades first, which is reasonable in this specific case.  If we had more code in production, I'd also add compatibility on the host side in a patch version of the previous release first.


src/messages/ClientToHost.ts
```diff
@@ -33,7 +33,8 @@ export function validate(msg: any): ClientToHost {
       registeredKeyFired: keyDownDecoder,
       client_started: startedDecoder,
       navRequest: navRequestDecoder,
-      toastRequest: toastDecoder
+      notifyRequest: notifyDecoder,
+      toastRequest: notifyDecoder
```
MattCheely: partial support for the old msgType value.

src/messages/Notification.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/** @external */
+const toastTypeDecoder: Decoder<'notifyRequest'> = map(
+  constant<'toastRequest'>('toastRequest'),
+  alwaysMsgType
+);
+
+/** @external */
+const decoder: Decoder<LabeledNotification> = object({
+  msgType: either(constant<'notifyRequest'>('notifyRequest'), toastTypeDecoder),
```
MattCheely: It's a bit hard to see what changed here with the file being renamed, but the `either` here combined with `toastTypeDecoder` and `alwaysMsgType` treats messages with `msgType: toastRequest` as valid and decodes them to a value with `msgType: notifyRequest`. 

This is the other really nice thing about decoders. In addition to detecting bad data at the edges of your system, they provide a well-defined location for handling compatibility when an API changes.

src/client.ts
```diff
@@ -288,28 +288,28 @@ bad input into one of the iframe-coordinator client methods.
...
    * @example
-   * `worker.requestToast({ title: 'Hello', message: 'World' });`
+   * `worker.requestnotification({ title: 'Hello', message: 'World' });`
    *
    * @example
-   * `worker.requestToast({ title: 'Hello', message: 'World', custom: { ttl: 5, level: 'info' } });`
+   * `worker.requestnotification({ title: 'Hello', message: 'World', custom: { ttl: 5, level: 'info' } });`
    */
-  public requestToast(toast: Toast): void {
+  public requestNotification(notification: Notification): void {
```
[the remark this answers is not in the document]
MattCheely: Now that I have added a changelog, yes.

cli/embedded-app/src/views/IframeEmbed.vue
```diff
@@ -9,7 +9,7 @@
     <frame-router
       id="frameRouter"
       v-bind:route="frameRoute"
-      v-on:toastRequest="displayToast"
+      v-on:notifyRequest="displayToast"
```
JackieChiles: I know it's an internal method, but could rename `displayToast` as well

src/client.ts
```diff
@@ -288,28 +288,28 @@ bad input into one of the iframe-coordinator client methods.
...
+   * presenting/styling the notification.  Application-specific concerns such as level, TTLs,
+   * ids for action callbacks (notification click, notification action buttons), etc. can be passed via
+   * the `custom` property of the `notification` type.
    *
-   * @param toast the desired toast configuration.
+   * @param notification the desired notification configuration.
    *
    * @example
-   * `worker.requestToast({ title: 'Hello world' });`
+   * `worker.requestnotification({ title: 'Hello world' });`
```
JackieChiles: Capital `N`?

cli/embedded-app/src/views/IframeEmbed.vue
MattCheely: I think I am going to leave these for now. I think another round of search and replace just makes it more likely a typo or error like some of the others y'all caught will slip in, and this is blocking the next release, which has some things other folks are waiting on.

PR #76, opened 2019-10-25
Fix issue with build not generating .d.ts type declarations.

MattCheely:
This was apparently the result of using the requireJsonModule flag to
enable import of package.json to get the library version. It didn't get
caught locally because previous builds had already generated the type
files.


MattCheely:
I'm going to preemptively merge this to unblock the build, but if anybody knows a better fix for this issue, please let me know.

PR #78, opened 2019-11-07
Add allow-modals to iframe sandbox

JackieChiles:
Add allow-modals to iframe sandbox to allow for native browser modals (confirm, alert, prompt).


MattCheely:
We still need a manual version bump for iframe-coordinator releases. If you can add that, and I'll merge it as soon as it's in.

PR #85, opened 2019-11-12
Comui 167

MattCheely:
Just a small tweak to the docs for this one.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -211,7 +211,13 @@ export class Client {
   }
 
   /**
-   * Gets the host url prefix for current app
+   * Translates a client route like `/foo/bar` to the full URL used in the host
+   * app for the same page, e.g. `https://hostapp.com/#/client-app/foo/bar`.
+   * You should use this whenver generating an internal link within a client
+   * application so that the user gets a nice experience if they open a link in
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -211,7 +211,13 @@ export class Client {
...
 
   /**
-   * Gets the host url prefix for current app
+   * Translates a client route like `/foo/bar` to the full URL used in the host
+   * app for the same page, e.g. `https://hostapp.com/#/client-app/foo/bar`.
+   * You should use this whenver generating an internal link within a client
+   * application so that the user gets a nice experience if they open a link in
+   * a new tab, or copy and paste a link URL into a chat message or email.
+   *
+   * @param clientRoute The /-separated path within the client app to link to.
```
JackieChiles: Wow, somehow I totally overlooked that this exists. Thanks!

MattCheely:
Found a couple of other minor improvements to make while reviewing the docs and added them.

src/client.ts
MattCheely: We have @zsdsjl to thank for the implementation! We didn't use it right away, and I forgot about it, because I have about 2k of working memory. Thanks again @zsdsjl !

PR #86, opened 2019-11-13
Opt in click handler

JackieChiles:
I bumped the major version because I'd consider removal of the automatic click listener a breaking change.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -35,6 +35,7 @@ export interface ClientConfigOptions {
  */
 export class Client {
   private _isStarted: boolean;
+  private _isInterceptingLinksStarted: boolean;
   private _clientWindow: Window;
   private _environmentData: EnvData;
   private _envDataEmitter: InternalEventEmitter<EnvData>;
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -259,11 +260,35 @@ bad input into one of the iframe-coordinator client methods.
     this._isStarted = true;
 
     this._clientWindow.addEventListener('message', this._onWindowMessage);
-    this._clientWindow.addEventListener('click', this._onWindowClick);
```
stormojm: Why not leave the default behavior intact and provide an option to start for deciding wether or not to intercept the links?

Then we don't have to bump the major.

MattCheely: We could, but IMO the separate API call is nicer and it's pretty easy to do the major bump now while the library isn't widely used. Also, since it doesn't introduce client/host compatibility issues, it won't force anybody having success with an older version to upgrade.  That said, I'm open to the idea if I've missed a reason this will cause pain for somebody.

JackieChiles: Agreed, I'd rather do the major bump now. Happy to change though if needed.

PR #87, opened 2019-11-25
[QM-6811] allow form submissions in client frame by default

jakenewby:


MattCheely:
@jakenewby Can you run `npm version patch` and push that commit? The build doesn't auto-bump versions yet.

jakenewby:
^ done

src/FrameManager.ts
```diff
@@ -68,7 +68,7 @@ class FrameManager {
 
     this._iframe.setAttribute(
       'sandbox',
-      'allow-scripts allow-same-origin allow-modals'
+      'allow-scripts allow-same-origin allow-modals allow-forms'
```
justinaray: I think the allow-forms is pretty safe to add here unconditionally, so I'm fine with this.

It does bring up a broader question if some of the more risky (allow-scripts, allow-same-origin, allow-top-navigation, etc.) and more UX-affecting (allow-modals, allow-popups, etc.) switches should be opt-in.  We could allow the host to specify these or try to make educated decisions based on if the iframe is from the same origin or not.

jakenewby: agreed. I think, optimally, we would provide configuration options per-client with some sane fallbacks (just "allow-scripts", probably).

PR #89, opened 2020-02-04
add allow-popups to iframe sandbox attribute list

jakenewby:


src/FrameManager.ts
```diff
@@ -68,7 +68,7 @@ class FrameManager {
 
     this._iframe.setAttribute(
       'sandbox',
-      'allow-scripts allow-same-origin allow-modals allow-forms'
+      'allow-scripts allow-same-origin allow-modals allow-forms  allow-popups'
```
JackieChiles: Super minor but looks like an extra space here

jakenewby: might have been the merge conflict resolution that got me there.  thanks

MattCheely:
You'll need to bump the patch version to get it to publish correctly. We still haven't gotten around to doing that in CI yet.

PR #90, opened 2020-02-06
COMUI-184 feature(ifc-cli) 

djhouseknecht:
Added option to ifc-cli to serve bare-bones host app over ssl


.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules/
 /dist/
 /cli/embedded-app/dist/**/*
 doc/
+.vscode
```
Thiyagarajan-A: Add a new line at the end, just to make the git happier on processing of CR and CRLF.

README.md
```diff
@@ -165,6 +165,7 @@ Usage: ifc-cli [options]
 Options:
   -f, --config-file <file>  iframe client configuration file (default: "/home/mcheely/projects/iframe-coordinator/ifc-cli.config.js")
   -p, --port <port_num>     port number to host on (default: 3000)
+	-s, --ssl                 serve over https
```
Thiyagarajan-A: Remove extra spaces at the start. Align/indent  properly.

djhouseknecht:
@Thiyagarajan-A I corrected the formatting issues and added an `.editorconfig` file to prevent that mistake from happening in the future. Feel free to adjust its contents.

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.{diff,md}]
+trim_trailing_whitespace = false
```
MattCheely: We have formatting configured with prettier that runs on commit, so I don't think we want to add this, at least not the `*` rule.  Otherwise we have two sources of truth for formatting that can become in conflict with each other.  If you want to enforce formatting rules for markdown, update the `lint-staged` rule in `package.json` accordingly

djhouseknecht: I don't want to clutter up config if you already have prettier. I will push a commit to remove this file.

PR #92, opened 2020-02-13
Default to about:blank when changing route

shanebruggeman:
Default new path to render to about:blank when changeRoute is called with a bad / falsey value

### DIFF
--- a/src/HostRouter.ts
+++ b/src/HostRouter.ts
@@ -25,14 +25,10 @@ export class HostRouter {
    * @param route The route to lookup, such as '/foo/bar/baz'
    */
   public getClientTarget(route: string): ClientTarget | null {
-    const desiredRoute = route || '';
     let clientTarget: ClientTarget | null = null;
 
     this._clients.forEach(client => {
-      const clientRoute = matchAndStripPrefix(
-        desiredRoute,
-        client.assignedRoute
-      );
+      const clientRoute = matchAndStripPrefix(route, client.assignedRoute);
       if (clientRoute !== null) {
         clientTarget = {
           id: client.id,
--- a/src/HostRouter.ts
+++ b/src/HostRouter.ts
@@ -26,7 +26,6 @@ export class HostRouter {
    */
   public getClientTarget(route: string): ClientTarget | null {
     const desiredRoute = route || '';
-
     let clientTarget: ClientTarget | null = null;
 
     this._clients.forEach(client => {
--- a/src/HostRouter.ts
+++ b/src/HostRouter.ts
@@ -24,10 +24,8 @@ export class HostRouter {
    *
    * @param route The route to lookup, such as '/foo/bar/baz'
    */
-  public getClientTarget(route: string): ClientTarget {
-    if (!route) {
-      return null;
-    }
+  public getClientTarget(route: string): ClientTarget | null {
+    const desiredRoute = String(route || '');
 
     let clientTarget: ClientTarget = {
       id: null,
@@ -36,7 +34,10 @@ export class HostRouter {
     };
 
     this._clients.forEach(client => {
-      const clientRoute = matchAndStripPrefix(route, client.assignedRoute);
+      const clientRoute = matchAndStripPrefix(
+        desiredRoute,
+        client.assignedRoute
+      );
       if (clientRoute !== null) {
         clientTarget = {
           id: client.id,
@@ -46,7 +47,7 @@ export class HostRouter {
       }
     });
 
-    return clientTarget;
+    return clientTarget.id ? clientTarget : null;
   }
 }
 
--- a/src/HostRouter.ts
+++ b/src/HostRouter.ts
@@ -25,13 +25,9 @@ export class HostRouter {
    * @param route The route to lookup, such as '/foo/bar/baz'
    */
   public getClientTarget(route: string): ClientTarget | null {
-    const desiredRoute = String(route || '');
+    const desiredRoute = route || '';
 
-    let clientTarget: ClientTarget = {
-      id: null,
-      url: null,
-      assignedRoute: null
-    };
+    let clientTarget: ClientTarget | null = null;
 
     this._clients.forEach(client => {
       const clientRoute = matchAndStripPrefix(
@@ -47,7 +43,7 @@ export class HostRouter {
       }
     });
 
-    return clientTarget.id ? clientTarget : null;
+    return clientTarget;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/elements/frame-router.ts
```diff
@@ -102,6 +102,8 @@ export default class FrameRouterElement extends HTMLElement {
    * @param newPath a new route which matches those provided originally.
    */
   public changeRoute(newPath: string) {
+    newPath = newPath || 'about:blank';
```
MattCheely: This doesn't really make sense, because we don't want to `getClientTarget` on `about:blank`.  As currently written, the idea is that if `newPath` is invalid in any way `clientInfo` should be null, and `setFrameLocation` on line 117/119 should be called with `null` and then `setFrameLocation` sets `about:blank` ([see here](https://github.com/purecloudlabs/iframe-coordinator/blob/master/src/FrameManager.ts#L85)).  

I think if we are trying to fall-back to `about:blank` in multiple locations, it makes things harder to understand, and opens us up to more potential bugs in the future. Can you figure out why the current mechanism isn't working?

shanebruggeman: The current mechanism isn't working because we expect the passed value to be a string and not null / undefined / etc and our string processing tools break trying to use string methods (`matchAndStripPrefix` being the primary offender).

An alternative fix would be to return the "not found" case from client target lookup if no route was passed
```
let clientTarget: ClientTarget = {
      id: null,
      url: null,
      assignedRoute: null
};
```

Then the frame router (here) would behave as if no viable client was found and default to about:blank without any further changes. Would that be reasonable?

MattCheely: Looking at the code a bit more, I think this may stem from bad type signatures in `attributeChangedCallback` where we don't account for attributes possibly being null. We should update that type signature accordingly and follow it through.   

That generally means making the changes you suggest but instead of having `getClientTarget` return an object full of nulls when no client is found, we should probably just return null. That way we avoid the weird invalid state ambiguity of what it means if some properties are null and others are not.

src/HostRouter.ts
```diff
@@ -24,14 +24,20 @@ export class HostRouter {
    *
    * @param route The route to lookup, such as '/foo/bar/baz'
    */
-  public getClientTarget(route: string): ClientTarget {
+  public getClientTarget(route: string): ClientTarget | null {
+    const desiredRoute = String(route || '');
```
MattCheely: I don't think you need the string constructor here? `route || ''` should suffice.

src/HostRouter.ts
```diff
@@ -41,7 +47,7 @@ export class HostRouter {
       }
     });
 
-    return clientTarget;
+    return clientTarget.id ? clientTarget : null;
```
MattCheely: This is a weird way to do this. I would just start with `clientTarget` initialized to null and leave the return statement as is.

src/elements/frame-router.ts
```diff
@@ -129,11 +130,11 @@ export default class FrameRouterElement extends HTMLElement {
    */
   public attributeChangedCallback(
     name: string,
-    oldValue: string,
-    newValue: string
+    oldValue: unknown,
+    newValue: unknown
```
MattCheely: `unknown` is not the correct type here. I think `string | null` is accurate, or at least accurate enough.  I don't think browsers will pass undefined here. If you wanted to account for that, something like `oldValue?: string | null` would work fine, although I really don't think it's necessary.  `unknown` indicates we have no idea what it is, could be a number or an object or a function, which doesn't really apply here.

src/elements/frame-router.ts
```diff
@@ -73,7 +73,8 @@ export default class FrameRouterElement extends HTMLElement {
       hostRootUrl: processedHostUrl
     };
 
-    this.changeRoute(this.getAttribute(ROUTE_ATTR) || 'about:blank');
+    const frameRoute = String(this.getAttribute(ROUTE_ATTR) || '');
```
MattCheely: There's not really a good reason to add these empty string fallbacks everywhere _and_ update `getClientTarget` to handle null inputs. What I was getting at attempting is to let the null value of the attribute continue through the call stack, and when it comes time to actually set the frame location , it should fall back to `about:blank` when presented null client data.

src/elements/frame-router.ts
```diff
@@ -129,11 +129,11 @@ export default class FrameRouterElement extends HTMLElement {
...
   public attributeChangedCallback(
     name: string,
-    oldValue: string,
-    newValue: string
+    oldValue: string | null,
+    newValue: string | null
   ) {
     if (name === ROUTE_ATTR && oldValue !== newValue) {
-      this.changeRoute(newValue);
+      this.changeRoute(String(newValue || ''));
```
MattCheely: You don't need the `String(...)` constructor here.  I probably would have let the null value cascade through, but this will also work and isn't hard to change if we need to in the future.

src/HostRouter.ts
```diff
@@ -24,14 +24,15 @@ export class HostRouter {
...
    * @param route The route to lookup, such as '/foo/bar/baz'
    */
-  public getClientTarget(route: string): ClientTarget {
-    let clientTarget: ClientTarget = {
-      id: null,
-      url: null,
-      assignedRoute: null
-    };
+  public getClientTarget(route: string): ClientTarget | null {
+    const desiredRoute = route || '';
```
MattCheely: With the compiler guarantee that `route` is always a string, you no longer need the intermediate `desiredRoute` variable for handling null values.

MattCheely:
You will need to rebase off of master - there are some type errors that will show up at that point that need to be fixed around David's recent changes.

src/elements/frame-router.ts
```diff
@@ -118,8 +118,10 @@ export default class FrameRouterElement extends HTMLElement {
         clientInfo && clientInfo.url
       );
 
-      this._frameManager.setFrameSandbox(clientInfo.sandbox);
-      this._frameManager.setFrameAllow(clientInfo.allow);
+      if (clientInfo) {
+        this._frameManager.setFrameSandbox(clientInfo.sandbox);
+        this._frameManager.setFrameAllow(clientInfo.allow);
+      }
```
shanebruggeman: @djhouseknecht clientInfo moves to possibly-null with these changes. I added a conditional around these but I'm rethinking this and wondering if maybe we need to set `sandbox`/`allow` off in the case we can't find clientInfo. Any thoughts?

shanebruggeman: The more I think of it, the more it makes sense to 1) keep the behavior the same and 2) ensure that `sandbox / allow` changes when the route does. So removed the conditional and just pass undefined if there's no clientInfo

djhouseknecht: Yeah, that is exactly the thinking behind that. I was about to comment and then saw your commit. The new null check looks good 👍

PR #93, opened 2020-02-21
Ifc cli not found

MattCheely:
I'm working on some quality-of-life improvements for ifc-cli and this is the first one. It updates the host app to show some useful details when you navigate to a route with no assigned clients. Users will either see a list of registered client apps (with links) or a message saying no clients have been registered.

### DIFF
--- a/cli/embedded-app/src/views/IframeEmbed.vue
+++ b/cli/embedded-app/src/views/IframeEmbed.vue
@@ -6,6 +6,23 @@
       as
       <span class="frame-url">{{ frameUrl }}</span>
     </div>
+    <div v-if="showMenu" id="appMenu">
+      <h2>No app is registered for {{ frameRoute }}</h2>
+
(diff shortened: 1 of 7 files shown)


cli/embedded-app/src/views/IframeEmbed.vue
```diff
@@ -71,10 +91,17 @@ export default {
...
+      this.showMenu = this.frameUrl === 'about:blank';
     }
   },
   mounted() {
     // Call the custom config set up on the CLI.
+    const oldSetupFrames = frameRouter.setupFrames;
+    frameRouter.setupFrames = (...args) => {
+      this.clientConfig = args[0];
+      oldSetupFrames.apply(frameRouter, args);
+    };
```
MattCheely: The function proxying here is a bit clunky, but I don't want to expose the frame-router's internal client config publicly, and since this is part of the same repo/project as the frame router element itself, I think it's more reasonable than it would be otherwise.

PR #94, opened 2020-03-11
Add proxy route for same-domain host and client hosting.

MattCheely:
This adds a built-in proxy route to `ifc-cli` so that the built in host app and local clients can live on the same domain from the browser's perspective, allowing shared localstorage and other domain-specific APIs.


MattCheely:
Looks like my editor's prettier config updated some html/README files, but most of the proxy-relevant changes should be pretty clear, I think.

PR #106, opened 2020-07-09
Preapare for npmjs publishing

MattCheely:
This primarily just changes the build process and publication target, but I've also added the standard commit format from other common UI repos to get us better change logs for public consumption.


MattCheely:
I will probably need to try building from this branch before merging to iron out the inevitable CI snafus

PR #111, opened 2020-07-27
add support for nested iframe-coordinators

MikeBlandford:
client to host and host to client messages are marked so that the wrong target does not accept them

can be tested using the following package.json entry (on the Publish branch version because dist files were added) 

`"iframe-coordinator": "git+https://github.com/MikeBlandford/iframe-coordinator.git#nestedIframesPublish"
`

### DIFF
--- a/src/FrameManager.ts
+++ b/src/FrameManager.ts
@@ -6,7 +6,11 @@ import {
   HostToClient,
   validate as validateOutgoing
 } from './messages/HostToClient';
-import { API_PROTOCOL, PartialMsg } from './messages/LabeledMsg';
+import {
+  API_PROTOCOL,
+  applyHostProtocol,
+  PartialMsg
+} from './messages/LabeledMsg';
 
 /** @external */
 const IFRAME_STYLE = `
@@ -143,16 +147,14 @@ class FrameManager {
    *
    * @param message The message to send.
    */
-  public sendToClient<T, V>(message: PartialMsg<T, V>) {
+  public sendToClient<T, V>(partialMsg: PartialMsg<T, V>) {
     const clientOrigin = this._expectedClientOrigin();
     if (this._iframe.contentWindow && clientOrigin) {
+      const message = applyHostProtocol(partialMsg);
       let validated: HostToClient;
 
       try {
         validated = validateOutgoing(message);
-        if (validated.direction === undefined) {
-          validated.direction = 'HostToClient';
-        }
       } catch (e) {
         throw new Error(
           `
--- a/src/FrameManager.ts
+++ b/src/FrameManager.ts
@@ -2,7 +2,10 @@ import {
   ClientToHost,
   validate as validateIncoming
 } from './messages/ClientToHost';
-import { validate as validateOutgoing } from './messages/HostToClient';
+import {
+  HostToClient,
+  validate as validateOutgoing
+} from './messages/HostToClient';
 import { API_PROTOCOL, PartialMsg } from './messages/LabeledMsg';
 
 /** @external */
@@ -143,10 +146,13 @@ class FrameManager {
   public sendToClient<T, V>(message: PartialMsg<T, V>) {
     const clientOrigin = this._expectedClientOrigin();
     if (this._iframe.contentWindow && clientOrigin) {
-      let validated = null;
+      let validated: HostToClient;
 
       try {
         validated = validateOutgoing(message);
+        if (validated.direction === undefined) {
+          validated.direction = 'HostToClient';
+        }
       } catch (e) {
         throw new Error(
           `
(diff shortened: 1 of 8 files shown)


src/messages/LabeledMsg.ts
```diff
@@ -35,6 +42,8 @@ export interface LabeledMsg<T, V> extends PartialMsg<T, V> {
   protocol: 'iframe-coordinator';
   /** library version */
   version: string;
+  /** So that nested iframe-coordinators can ignore messages that don't apply */
+  direction?: MessageDirection;
```
MattCheely, line 46: Are we sure this actually needs to be optional? From the perspective of typechecking in this library it should always be set.

src/FrameManager.ts
```diff
@@ -143,10 +146,13 @@ class FrameManager {
   public sendToClient<T, V>(message: PartialMsg<T, V>) {
     const clientOrigin = this._expectedClientOrigin();
     if (this._iframe.contentWindow && clientOrigin) {
-      let validated = null;
+      let validated: HostToClient;
 
       try {
         validated = validateOutgoing(message);
+        if (validated.direction === undefined) {
+          validated.direction = 'HostToClient';
```
MattCheely, line 154: I'm fairly certain we don't need/want the conditional here. Every outgoing message from the host should be tagged as `HostToClient`. Only setting if it's undefined implies that the app interfacing with this might somehow be producing a message that has `direction` already set, which is probably not a good idea.

src/client.ts
```diff
@@ -231,10 +235,14 @@ export class Client {
...
     const message = applyProtocol(partialMsg);
-    let validated = null;
+
+    let validated: ClientToHost;
 
     try {
       validated = validateOutgoing(message);
+      if (validated.direction === undefined) {
+        validated.direction = 'ClientToHost';
+      }
```
MattCheely, line 245: Same comment as on the host side: I don't think we need/want the conditional. Everything outbound from here should always be flagged as `ClientToHost`

src/messages/LabeledMsg.ts
```diff
@@ -57,7 +66,12 @@ export interface PartialMsg<T, V> {
 export function applyProtocol<T, V>(
   partialMsg: PartialMsg<T, V>
 ): LabeledMsg<T, V> {
-  return { ...partialMsg, protocol: API_PROTOCOL, version };
+  return {
+    direction: undefined,
```
MattCheely, line 70: My initial inclication here would be to require a direction when calling `applyProtocol` rather than leaving it undefined. Or perhaps to create separate `applyHostProtocol` and `applyClientProtocol` messages

src/messages/LabeledMsg.ts
MikeBlandford: With a host and client, and only one of them has updated iframe-coordinator, I believe it would validate incoming messages from the non-upgraded version and say they are invalid if this was not optional

MikeBlandford: Oh you mean the type, not the validation, which happens in directionDecoder. I'll try that.

src/messages/LabeledMsg.ts
MikeBlandford: I tried that at first but then realized that applyProtocol is only used by the client. Oh, I can rename it to applyClientProtocol and then have the host do its own thing.

MikeBlandford: Oh, maybe it should be done on host too, since host is currently setting version to 'unknown' through the decoders

MikeBlandford: I changed it to use the protocols instead of manually setting direction, and it sets version now too.

src/messages/LabeledMsg.ts
MattCheely: Ah, the decoder needing it makes sense. I was thinking that it wouldn't matter with old clients since at runtime all the type information is gone anyway, but the decoder definitely needs to handle the undefined case. Thanks for clarifying!

MattCheely:
There is one remaining issue, which is that while you were working on this change we added some commit message linting to the repo in master (which helps us generate changelogs and maintain semver), so your commit messages cause linting to fail. The fix should be fairly easy if you rebase into a single commit onto master and update the commit message like this:

```
feat(messaging): support nested iframes by tagging messages with direction
 
Cross-frame messages are now tagged as `HostToClient` or `ClientToHost` so that a frame running both
host and client code will not try to incorrectly process messages intended for the host code with
the client code and vice-versa.
```

there's an `npm run commit` command to help walk through generating messages, but I don't think it works with rebase, so I generated the above for you.  You can validate that it worked by running `npm run lint`

MikeBlandford:
Ok. I'll close this and open a new pull request for the rebase: https://github.com/purecloudlabs/iframe-coordinator/pull/114

PR #126, opened 2021-02-05
fix(cli): remove embedded app dual scrollbars

farooqu:
* Give html height of 100%
* Remove positioning of frame-router and iframe
* This will make only the contents of the iframe scroll as needed


farooqu:
Before:
Before

After:
After

MattCheely (approved):

PR #137, opened 2021-03-31
Build improvements

MattCheely:
CI failed on the last merge in Jenkins, so I'm taking an opportunity while fixing it to bring over some of the general CI improvements we've made in the web component repo.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,22 @@
+@Library('pipeline-library')
+import com.genesys.jenkins.Service
+
+def notifications = null
+String[] mailingList = [
+  "Matthew.Cheely@genesys.com",
+  "Daragh.King@genesys.com"
+]
+
+def isReleaseBranch() {
+    return env.SHORT_BRANCH.equals('master');
+}
+
 pipeline {
   agent { label 'dev_mesos_v2' }
+  options {
+    quietPeriod(480)
+    disableConcurrentBuilds()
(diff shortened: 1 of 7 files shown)


Jenkinsfile
```diff
@@ -49,17 +69,15 @@ pipeline {
...
     stage('Publish Library') {
+      when {
+        expression { isReleaseBranch()  }
+      }
       steps {
         dir(env.REPO_DIR) {
-          sh '''
-             echo "registry=https://registry.npmjs.org" > ./.npmrc
-             echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> ./.npmrc
-          '''
```
MattCheely, line 57: I believe this is unnecessary after removing the `.npmrc` file from the repo and adding a publishConfig to package.json. Also, I really don't want to end up accidentally committing credentials so it's best to just not put them on disk in the project.

Jenkinsfile
```diff
@@ -1,5 +1,22 @@
+@Library('pipeline-library')
+import com.genesys.jenkins.Service
+
+def notifications = null
+String[] mailingList = [
+  "Matthew.Cheely@genesys.com",
+  "Daragh.King@genesys.com"
```
daragh-king-genesys, line 7: Would a team mailing list be easier to manage?

daragh-king-genesys (approved):

PR #138, opened 2021-04-01
[NO-JIRA] sets ifc-cli.config.js to esm module

liropa:
This is a breaking change.

This updates the ifc-cli.config.js as an esm module, allowing for imports. The test/sample config files and the README were updated to reflect the change.

There was also a dependency issue when running the tests that required a major package bump (karma-webpack). After minor karma.config update due to the major bump/breaking change, tests where running successfully and all passing.


liropa:
Declining

PR #140, opened 2021-04-02
More Jenkins cleanup

MattCheely:
- Add step to avoid build loops that I missed previously
- Move embedded app installs to a different npm lifecycle stage so they don't break npm publish

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "lint.format": "prettier --check **/*.ts **/*.tsx",
     "lint.commit": "commitlint -f e307dddc51545900940861f06baf210cd4a79cec",
     "lint": "tslint --project tsconfig.json && npm run lint.format && npm run lint.commit",
-    "prepare": "./scripts/prepare-deps.sh",
+    "preinstall": "./scripts/prepare-deps.sh",
     "release": "standard-version",
     "start": "webpack --watch",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
     "lint.format": "prettier --check **/*.ts **/*.tsx",
     "lint.commit": "commitlint -f e307dddc51545900940861f06baf210cd4a79cec",
     "lint": "tslint --project tsconfig.json && npm run lint.format && npm run lint.commit",
-    "prepare": "./scripts/prepare-deps.sh",
+    "preinstall": "./scripts/prepare-deps.sh",
```
MattCheely, line 71: I have a vague fear this may run on downstream installs and cause problems, but the npm docs are not clear, and it should at least let us start building again to find out.

JackieChiles: It was true in the past, at least, that these scripts would run for downstream consumers, though I'm not sure if that's still the case as running unknown code from third party dependencies on `npm install` is of course not ideal for consumers. I believe the `--ignore-scripts` flag will skip those scripts on dependencies, at least.

JackieChiles (approved):

package.json
MattCheely: I decided to just remove this entirely for now. It's just a convenient helper for local development. I can come back to find a better solution later.

PR #146, opened 2021-05-04
feat(ifc-api-improvements): Improve URL and link generation APIs

katie-bobbe-genesys:


client-app-example/public/client-app-1/index.html
```diff
@@ -8,21 +8,37 @@ <h1>Application #1</h1>
     <h2>Routing</h2>
     <h3>Current Local Route:</h3>
     <div id="path"></div>
-    <h3>Current Host Route:</h3>
-    <div id="hostPath"></div>
+    <h3>Current Host Route Calculated From Client:</h3>
```
katie-bobbe-genesys, line 11: I will go back and clean up some of these client app examples so it better showcases how to generate links, especially using the new custom elements `ifc-client-link` and `ifc-host-link`

src/elements/ifc-client-link.ts
```diff
@@ -0,0 +1,58 @@
...
+  private _environmentData: EnvData;
+  private _assignedRoute: string;
+
+  /** @internal */
+  constructor() {
+    super();
+    this._client = new Client({
+      hostOrigin: 'http://localhost:8080'
+    });
+    this._environmentData = {
```
katie-bobbe-genesys, line 24: I don't know how to get the `environmentalData` and `assignedRoute` to the component. Here I just hard coded some values.

src/elements/ifc-host-link.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export default class IfcHostLinkElement extends HTMLElement {
+  private _client: Client;
+  private _environmentData: EnvData;
+
+  /** @internal */
+  constructor() {
+    super();
+    this._client = new Client();
+    this._environmentData = {
```
katie-bobbe-genesys, line 21: Same as with the `ifc-client-link` component above, I don't know how to get the `environmentData` to the component. Here I just hard coded some values.

src/index.ts
```diff
@@ -1,4 +1,9 @@
-import { Client } from './client';
+import { Client, registerCustomElementsClient } from './client';
```
katie-bobbe-genesys, line 1: I am registering custom elements from both the Client and the Host. Is there a better way to do this?

katie-bobbe-genesys:
Here is my WIP for [COMUI-485](https://inindca.atlassian.net/browse/COMUI-485). I have left comments on the specific areas that I have questions about. I would also just like general feedback on the overall approach that I took here, and make sure it's on the right track. Thank you.

src/client.ts
```diff
@@ -19,19 +21,28 @@ import {
...
 import { Notification } from './messages/Notification';
 import { Publication } from './messages/Publication';
 
+/**
+ * Registers custom elements used by the client application
+ */
+function registerCustomElementsClient(): void {
+  customElements.define('ifc-client-link', IfcClientLinkElement);
+  customElements.define('ifc-host-link', IfcHostLinkElement);
+}
```
MattCheely, line 41: Instead of making `registerCustomElementsClient` a standalone function, we can make a `registerCustomElements` method on the `Client` class.  I think this is key to solving most of the issues you're dealing with.  It moots the export issues because there's no new export, and now you have access to a client instance when registering the custom elements.

I'd then update the IfcClientLinkElement to take a client instance as a constructor argument and in the 
`registerCustomElements` method, extend that class, passing the client instance via `super`. Something like this, but maybe with better class names: 

```typescript
registerCustomElements(): void {
  const clientInstance = this;
  class IfcClintLinkElementComplete extends IfcClientLinkElement {
     constructor() {
         super(clientInstance);
     }
  }

 customElements.define('ifc-client-link', IfcClientLinkElementComplete);
}
```

src/elements/ifc-client-link.ts
MattCheely: With access to a `Client` instance, you won't need the environment data explicitly, since you can just use `urlFromClientPath` (which you are doing already!)

src/elements/ifc-client-link.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /**
+   * @internal
+   * @inheritdoc
+   */
+  public connectedCallback() {
+    const thisInnerHTML = this.innerHTML;
+    let path = this.getAttribute(PATH_ATTR) || 'about:blank';
+    const urlFromClientPath = this._client.urlFromClientPath.bind(this);
```
MattCheely, line 54: You shouldn't need the `bind(this)` for this call, at least not once you're able to pass in an initialized client instance. You can just directly call `this._client.urlFromClientPath(path)`

src/elements/ifc-client-link.ts
```diff
@@ -0,0 +1,58 @@
...
+    path = urlFromClientPath(path);
```
MattCheely, line 55: There is a (probably rare) potential issue to think about which is that the client doesn't always have the environment data right away, since all messaging with the host app is async.  While unlikely, it's possible that an app might render this component before the environment data arrives, but that's easy enough to work around. In this function, I'd also add a [listener on the client](https://apps.inindca.com/common-ui-docs/iframe-coordinator/classes/client.client-1.html#addlistener) for changes to environment data and update the link path whenever the listener is triggered.

MattCheely (changes requested):

src/client.ts
```diff
@@ -94,6 +85,27 @@ export class Client {
     this._assignedRoute = null;
   }
 
+  /**
+   * Registers custom elements used by the client application
+   */
+  public registerCustomElements(): void {
+    const clientInstance = this;
+    class IfcClientLinkElementComplete extends IfcClientLinkElement {
```
katie-bobbe-genesys, line 93: @MattCheely Changing the `registerCustomElements` standalone function to a method on the `Client` class solved the issues I was facing with being unable to access the environmentData. But when I tried to commit my changes, I got a linting error `A maximum of 1 class per file is allowed.`. Did I implement this incorrectly?

src/client.ts
```diff
@@ -94,6 +85,27 @@ export class Client {
...
+      constructor() {
+        super(clientInstance);
+      }
+    }
+    class IfcHostLinkElementComplete extends IfcHostLinkElement {
+      constructor() {
+        super(clientInstance);
+      }
+    }
+    clientInstance.addListener('environmentalData', envData => {
```
katie-bobbe-genesys, line 103: @MattCheely I needed to add a listener to wait for the environmentalData before rendering the link components. I found that in the sample app, the link components would render before the environmentalData loads. Adding a listener here solves that problem, but I would like to know if this is a good solution or not.

src/client.ts
MattCheely: Not at all. Normal best practice is one class per file, but that's working under the assumption that you're making classes for external consumption, or that the classes are complex. This is a special case in that we need to provide a class that can close over a local value. 

I'd just add a [tslint disable comment](https://palantir.github.io/tslint/usage/rule-flags/) for that rule on the line it's erroring on, presumably this one.

src/client.ts
```diff
@@ -242,14 +265,44 @@ export class Client {
...
+    const trimmedHostRoute = stripLeadingSlashAndHashTag(hostRoute);
+    return joinRoutes(hostRootUrl, trimmedHostRoute);
+  }
+
+  /**
+   * Deprecated -- replaced with urlFromClientPath
+   * Translates a client route like `/foo/bar` to the full URL used in the host
+   * app for the same page, e.g. `https://hostapp.com/#/client-app/foo/bar`.
+   *
+   * @param clientRouteLegacy Deprecated. The /-separated path within the client app to link to.
```
MattCheely, line 301: We should add/use the [tsdoc @deprecated flag](https://tsdoc.org/pages/tags/deprecated/) here instead of the descriptions.

MattCheely (changes requested):

src/client.ts
MattCheely: This is an interesting approach that I hadn't considered. by delaying the definition of the elements until the envData is present, the custom elements on the page won't do anything until the envData is received. Likely that means that the text that would be in the link just shows up as plain unstlyed text until envData arrives.

That makes the links not clickable, but also could create a somewhat jarring styling flash when they convert to links after the envData loads.  I'll outline/suggest another approach on the custom element itself that's closer to my original idea for comparison.  I'm not sure right now which I think is the best approach, I'd need to think about it a bit.

src/elements/ifc-client-link.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * @internal
+   * @inheritdoc
+   */
+  public connectedCallback() {
+    const thisInnerHTML = this.innerHTML;
+    let path = this.getAttribute(PATH_ATTR) || 'about:blank';
+    path = this._client.urlFromClientPath(path);
+    this.innerHTML = `<a href=${path}>${thisInnerHTML}</a>`;
```
MattCheely, line 38: Rather than setting `innerHTML` in the `connectedCallback`, I think we should be using the shadow DOM and regular DOM APIs or a template to build up the structure.  I think that's normally done in the constructor, but it could be done here as well.

There are some good examples from [google](https://developers.google.com/web/fundamentals/web-components/customelements#shadowdom) and [mozilla](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) that should help with getting started.

MattCheely (changes requested):

src/elements/ifc-client-link.ts
katie-bobbe-genesys: I was able to solve the above comment by adding the `target=_top` attribute to the link and adding the `allow-top-navigation-by-user-activation` keyword to the iframe sandbox

client-app-example/public/client-app-1/index.html
```diff
@@ -8,21 +8,37 @@ <h1>Application #1</h1>
...
+        <a class="client-transform-link" href="#/route2">This client: /route2</a>
+      </li>
+    </ul>
+
+    <h3>
+      Using custom elements ifc-client-link and ifc-host-link to build full urls
+    </h3>
+    <ul>
+      <li>
+        <ifc-client-link path="#/route2">Creating a full link from the client path using ifc-client-link -- to Route 2</ifc-client-link>
```
MattCheely, line 37: Minor, but for this and the host example below, I'd drop the `#` from the path and just use `/route2` and `/app2`.  It's good that things still work correctly if a `#` does get thrown in there, but I think from the user's perspective it makes more sense to only provide the path, and not a `#` that may or may not be present in the final generated URL in that position.

src/elements/ifc-client-link.ts
```diff
@@ -0,0 +1,30 @@
...
+    const slot = document.createElement('slot');
+    link.appendChild(slot);
+    const shadowRoot = this.attachShadow({ mode: 'open' });
+    shadowRoot.appendChild(link);
+    this._client.addListener('environmentalData', envData => {
+      let path = this.getAttribute(PATH_ATTR) || 'about:blank';
+      path = this._client.urlFromClientPath(path);
+      shadowRoot.firstElementChild?.setAttribute('href', path);
+    });
+  }
```
MattCheely, line 29: I missed it before, but we'll want to add [attribute observation](https://developers.google.com/web/fundamentals/web-components/customelements#attrchanges) for the path attribute, and update the generated link accordingly when the attribute changes.  That should be able to use the same function as the environmentData listener, I think.

src/elements/ifc-client-link.ts
```diff
@@ -0,0 +1,30 @@
...
+  /** @internal */
+  constructor(clientInstance: any) {
+    super();
+    this._client = clientInstance;
+
+    const link = document.createElement('a');
+    link.setAttribute('target', '_top');
+    const slot = document.createElement('slot');
+    link.appendChild(slot);
+    const shadowRoot = this.attachShadow({ mode: 'open' });
```
MattCheely, line 22: Sadly, I think I have to backtrack on the shadow DOM usage here.  It prevents global `a` tag styles from styling these generated links. Not only that, but adding styles to `ifc-client-link` elements doesn't change the internal `a` style unless we explicitly style the `a` tag in this component with `inherit` for things like color and text-decoration, due to the browser's default styles.  But applying `inherit` means that the links come out by default with regular text styles which is also bad. 

Sigh... This is the most awkward part of webcomponents.

MattCheely (changes requested):

MattCheely:
Looks like the build is failing due to a bad commit message (although it's hard to see why, looking at it).  I'd just rebase and squash the pr-fix commits once you're ready to merge to fix it.

src/elements/ifc-client-link.ts
```diff
@@ -0,0 +1,53 @@
...
+    return [PATH_ATTR];
+  }
+
+  /**
+   * @internal
+   * @inheritdoc
+   */
+  public connectedCallback() {
+    const content = this.innerHTML;
+    let path = this.getAttribute(PATH_ATTR) || 'about:blank';
```
MattCheely, line 33: One last thing I'd do for this PR is switch the `'about:blank'` default values to just an empty string.  If we append `about:blank` to the client or host URL, we just end up with nonsense, and if we end up with `about:blank` on it's own, we might accidentally send the user to a blank page, which would be a confusing experience.

MattCheely (changes requested):

MattCheely (approved):

PR #159, opened 2021-07-01
Add global modal messaging

S-Dingley-Genesys:
So, this is the first pass of the requestModal.

It has a modalType and modalData - I created the enum for the modal types  for client apps to use.
This isn't enforced on the requestModal parameter though, wasn't sure how to make it so it must be of one of those enum values.

The modal data currently has no structure, it can be anything. Not sure if this can be done more elegantly than it has been.


src/messages/Modal.ts
```diff
@@ -0,0 +1,18 @@
+import { constant, Decoder, mixed } from 'decoders';
+import { labeledDecoder, LabeledMsg } from './LabeledMsg';
+
+/**
+ * A message used to publish a generic messages
+ * between the clients and the host application.
+ */
+export interface LabeledModal extends LabeledMsg<'modalRequest', any> {
```
MattCheely, line 8: This file should be called `ModalRequest` and should follow the same pattern as other messages, like `NavRequest`. That means this interface would be come `LabeledModalRequest` and should extend `LabeledMsg<'modalRequest', ModalRequest>`, where `ModalRequest` is a type defined and exported in this file that contains all data involved in requesting a modal from the host application.

Additionally, you'll need to add the `LabeledModalRequest` to the `ClientToHost` type in `/messages/ClientToHost.ts`.

src/messages/Modal.ts
```diff
@@ -0,0 +1,18 @@
...
+  /** Message identifier */
+  msgType: 'modalRequest';
+}
+
+const decoder: Decoder<LabeledModal> = labeledDecoder(
+  constant<'modalRequest'>('modalRequest'),
+  mixed
```
MattCheely, line 15: This should be something that validates that the data is an actual modalRequest so we don't accidentally pass bad data to a host application.  Probably just validating that the request's `modalId` field is a string.

MattCheely (changes requested):
Documentation will need to be added in the `client-setup.md` file under "[Requesting Host Actions](https://github.com/purecloudlabs/iframe-coordinator/blob/master/doc/client-setup.md#requesting-host-actions)" to indicate that `requestModal` is an option in addition to `requestNotificaton`

src/client.ts
```diff
@@ -66,6 +66,13 @@ export class Client {
   private _registeredKeys: KeyData[];
   private _assignedRoute: string | null;
 
+  /**
+   * Supported modal types, will be used in web-dir to know which modal to launch
+   */
+  public modalTypes = {
+    WEM_COACHING_UI: 'wemCoachingUi'
+  };
```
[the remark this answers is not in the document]
S-Dingley-Genesys: Removed this, apologies.
I thought that there would have to be some kind of enforcement of the allowed type of modal and wasn't sure if that was would fit anywhere else.

src/client.ts
```diff
@@ -420,6 +427,22 @@ bad input into one of the iframe-coordinator client methods.
...
+  /**
+   * Asks the host application to launch an instance of a global modal
+   *
+   * Currently the only available one is WEM Coaching UI
+   *
+   * @param modalType the type of modal to be used.
+   * @param modalData any data you wish to be sent to the modal, Conversation IDs etc.
+   *
+   */
+  public requestModal(modalType: any, modalData: any) {
```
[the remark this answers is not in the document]
S-Dingley-Genesys: Changed as requested.

src/messages/Modal.ts
S-Dingley-Genesys: Changed as requested.

src/messages/Modal.ts
S-Dingley-Genesys: Changed as requested.

S-Dingley-Genesys (commented):
Comments given have been addressed.
enum removed, message changed and formatted/validated correctly and valid documentation updated.

daragh-king-genesys:
I don't have a lot of context for this new feature so if my comments don't make sense just ignore.

I think we need more documentation on how this will be implemented in the client (and host) app. It is a bit vgue at the moment - `pass an id and data`.

Will the client app get a message back from the host app? that the modal was displayed/closed/interacted with in a certain way?

S-Dingley-Genesys:
> I don't have a lot of context for this new feature so if my comments don't make sense just ignore.
> 
> I think we need more documentation on how this will be implemented in the client (and host) app. It is a bit vgue at the moment - `pass an id and data`.
> 
> Will the client app get a message back from the host app? that the modal was displayed/closed/interacted with in a certain way?

Have added a bit more detail on how to use in both client/host.

src/client.ts
```diff
@@ -420,6 +422,22 @@ bad input into one of the iframe-coordinator client methods.
     });
   }
 
+  /**
+   * Asks the host application to open a modal dialog.
+   *
+   * This is identified by the ModalRequest's `modalId` property.
```
MattCheely, line 428: The "This" in this sentence is ambiguous, IMO. I'd make it more explicit, something like:  "The `modalId` property names the modal that should be displayed"

doc/client-setup.md
```diff
@@ -106,9 +106,33 @@ let externalLink = `<a href="https://external-site.com/external/path" target="_t
...
+});
+```
+The 'modalRequest' event is received by the host application.
+Below is an example (Ember).
+```javascript
+modalRequest(event) {
+  // Do something with the event here
+},
+this.element.addEventListener('modalRequest', this.modalRequest.bind(this));
+```
```
MattCheely, line 135: This host example is pulled from web-dir, but only makes sense with a lot of implicit context from that app that may not carry over in general.  I'd just summarize that the `frame-router` element will emit a custom event with the `ModalRequest` object in the detail property.

MattCheely (changes requested):

MattCheely:
> Will the client app get a message back from the host app? that the modal was displayed/closed/interacted with in a certain way?

I think we will want to add those sorts of events in the future, but they're not needed for Stephen's use case, so I'm fine with this minimal implementation for now.  This PR did remind me that we need a ticket to focus on and improve the host-side documentation in general, so I created one: https://inindca.atlassian.net/browse/COMUI-552

MattCheely (approved):

PR #168, opened 2021-12-03
feat(a11y): support adding a title attribute to the iframe

katie-bobbe-genesys:
COMUI-752

### DIFF
--- a/src/HostRouter.ts
+++ b/src/HostRouter.ts
@@ -34,7 +34,8 @@ export class HostRouter {
           url: applyRoute(client.url, clientRoute),
           assignedRoute: client.assignedRoute,
           allow: client.allow,
-          sandbox: client.sandbox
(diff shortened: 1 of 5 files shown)


src/HostRouter.ts
```diff
@@ -60,6 +61,8 @@ export interface ClientTarget {
   allow?: string;
   /** iframe's sandbox directive to be merged with defaults */
   sandbox?: string;
+  /** iframe's title attribute */
+  title?: string;
```
MattCheely, line 65: In the future, I think we'll want to allow for updating the title from the client app, so I would rename this to `defaultTitle`.  That way when we do allow clients to update, it's a little bit less confusing about how this property interacts with the client API for changing the title, and we don't have to introduce a breaking change or keep a legacy option under the old name in the future.

PR #203, opened 2022-08-24
iframe-coordinator monorepo setup

katie-bobbe-genesys:


package.json
```diff
@@ -1,93 +1,34 @@
...
-  "bin": {
-    "ifc-cli": "cli/ifc-cli.js"
+  "scripts": {
+    "build": "npm run build --workspaces --if-present",
+    "start-client-example": "npm run start --workspace=ifc-example-client",
+    "test": "npm test --workspaces --if-present",
+    "commit": "git-cz",
+    "lint": "commitlint -f de759fc509cf39895450dc31742cfb36b08e01b8",
+    "update-version": "./scripts/update-version.mjs",
+    "release": "npm run release --workspace=iframe-coordinator --workspace=ifc-cli",
```
MattCheely, line 16: The `update-version` and `release` commands seem to be conflicting in purpose here.  `release` is running `standard-version` in the library and cli packages independently, and `update-version` copies the top-level project version from this file to all of the child projects. 

I think if we run `standard-version` independently in each package, we run a risk of the versions getting out of sync. I would run `standard-version` in the project root to bump the package version in this file, and then run `update-version` (maybe should be called `sync-versions`?) to sync the root version to the child projects.  That would make the release script here something like:

`standard-version && npm run sync-versions`

where we might add some flags to standard-version based on beta status, etc.

package.json
```diff
@@ -1,93 +1,34 @@
...
+    "publish": "npm publish --workspace=iframe-coordinator --workspace=ifc-cli",
```
MattCheely, line 17: I think we'll need to change the `npm publish` command in the Jenkinsfile to be `npm run publish` rather than `npm publish` for this to work. I might even rename this to `publish-libs` just to avoid confusion of this script with the `npm publish` command.

MattCheely (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [5.0.1-beta.1](https://github.com/purecloudlabs/iframe-coordinator/compare/v5.0.2-beta.3...v5.0.1-beta.1) (2022-08-30)
```
MattCheely, line 5: This hasn't been published yet, so I think this is just a result of a local run of the version script. We should probably revert this, as we only want the changelog to update during the CI build where we publish the library.

package.json
```diff
@@ -1,93 +1,35 @@
...
-    "dist/index.js",
-    "dist/es5/client.js",
-    "dist/es5/host.js",
-    "dist/es5/index.js",
-    "dist/**/*.d.ts",
-    "cli/ifc-cli.js",
-    "cli/example-ifc.config.js",
-    "cli/embedded-app/dist/**/*"
+  "name": "iframe-coordinator-monorepo",
+  "version": "5.0.1-beta.1",
```
MattCheely, line 3: Similar to the changelog. I think this should stay 4.5.2 until the CI build, so that standard-version is working from the last published version.

package.json
```diff
@@ -1,93 +1,35 @@
...
-    "ifc-cli": "cli/ifc-cli.js"
+  "scripts": {
+    "build": "npm run build --workspaces --if-present",
+    "start-client-example": "npm run start --workspace=ifc-example-client",
+    "test": "npm test --workspaces --if-present",
+    "commit": "git-cz",
+    "lint": "commitlint -f de759fc509cf39895450dc31742cfb36b08e01b8",
+    "sync-versions": "./scripts/sync-versions.mjs",
+    "release": "standard-version --prerelease beta & npm run sync-versions",
+    "publish-libs": "npm publish --workspace=iframe-coordinator --workspace=ifc-cli",
```
MattCheely, line 17: let's add `--tag=beta` here until we're confident that everything is working and publishing correctly.  That way consumers of the library won't get the 5.x release right away when they just run `npm install iframe-coordinator`.

MattCheely (changes requested):

MattCheely:
Looks like github CI got caught in a loop running the tests. IIRC, that means we need to update the node version in the config.

MattCheely (approved):

PR #209, opened 2022-09-20
fix(cli): rename ifc-cli

katie-bobbe-genesys:
There is already an npm package named `ifc-cli`, so we renamed the package to `iframe-coordinator-cli`. Along the way we fixed build errors that were discovered during the process. To fix some of the build errors, we migrated to use `vite` in `iframe-coordinator-cli`.


packages/iframe-coordinator-cli/package.json
```diff
@@ -1,46 +1,34 @@
...
+    "vue": "^2.7.10",
     "vue-class-component": "^7.0.2",
     "vue-notification": "^1.3.16",
     "vue-property-decorator": "^8.1.0",
     "vue-router": "^3.0.3",
     "vuex": "^3.0.1"
   },
   "devDependencies": {
-    "@babel/preset-env": "^7.4.5",
-    "@types/chai": "^4.1.0",
```
katie-bobbe-genesys, line 25: I removed `types/chai`, `types/mocha` and `chai`. There were errors when testing caused by having duplicate identifiers with `jasmine`. (similar issue is described here https://github.com/angular/angular-cli/issues/3518) The types seemed to only be used in an example test, so I removed the example test and types.

katie-bobbe-genesys:
I tested this branch on Jenkins and the build passes successfully

PR #210, opened 2022-10-10
feat(frame-router): add option to setup frames via a property

katie-bobbe-genesys:
Add option to setup frames using a property

Being able set up the `frame-router` component via a property will make it easier to integrate with Elm.


packages/iframe-coordinator/src/elements/frame-router.ts
```diff
@@ -152,6 +152,26 @@ export default class FrameRouterElement extends HTMLElement {
     }
   }
 
+  /**
+   * @property {object} frameSetup A property that can be set to initialize the host frame with the
```
katie-bobbe-genesys, line 156: I think I'm missing something, because I'm getting a lint error `Documentation must exist for properties.` Any idea what to add here?

thomas-c-dillon: This is what I would do (that seems to address the linting concerns). I would create and interface for this object (and include the JSDoc comments there) and then add a simplified JSDoc comment on either the getter or setter.

Maybe something like this:
`/** A property that can be set to initialize the host frame with the
 * possible clients and the environmental data required by the clients
 */
interface FrameSetup {
  /** The map of registrations for the available clients. */
  clients: RoutingMap;
  /** Information about the host environment. */
  envData: EnvData;
}`

` private _frameSetup: FrameSetup;
  /** Additional Comments Here? */
  get frameSetup(): any {
    return this._frameSetup;
  }

  set frameSetup(frameSetup: FrameSetup) {
    this._frameSetup = frameSetup;
    this._setSetupFrames(frameSetup.clients, frameSetup.envData);
  }`

packages/iframe-coordinator/src/elements/frame-router.ts
```diff
@@ -152,6 +162,24 @@ export default class FrameRouterElement extends HTMLElement {
     }
   }
 
+  private _frameSetup: FrameSetup;
+
+  /**
+   * A property that can be set to initialize the host frame
+   */
+  get frameSetup(): any {
```
MattCheely, line 170: `frameSetup` feels like a bit of an awkward name for a property, although I can see the motivation in mirroring the method name. In hindsight, the method name isn't very good, as we're not really setting up the "frames", we're configuring the client routing and init. Also there's only actually one frame.  I would call this option something like `clientConfiguration` or `clientConfig`.  Maybe we even mark `setupFrames()` as deprecated when the 5.0 release goes out of beta.

packages/iframe-coordinator/src/elements/frame-router.ts
```diff
@@ -68,6 +78,9 @@ export default class FrameRouterElement extends HTMLElement {
    *
    * @param clients The map of registrations for the available clients.
    * @param envData Information about the host environment.
+   *
+   * @deprecated Use the new {@configureClients} method instead
```
katie-bobbe-genesys, line 82: @MattCheely in a previous comment you mentioned that we can deprecate `setupFrames()`. I added a new method `configureClients()` that can replace `setupFrames()`. Is this what you were thinking? Or were you thinking that we would remove the public method all together in favor of using the property?

MattCheely: That is a good question. 🤔 I would say that for now, we can just deprecate `setupFrames()` without adding a replacement function. We'll announce the deprecation and get feedback from consumers to see if moving to a property causes any problems for people, and then create the new function if it does. That way, if the property is good enough on it's own there will only be one way to do it. That's usually preferable.

MattCheely (changes requested):
Thanks for taking on the deprecation as well!

MattCheely (approved):

PR #214, opened 2022-11-18
build(bundle): switch bundler from webpack to rollup

MattCheely:
This should improve compatibility with some of our apps.

I tested it linked locally with web-dir and the nav-system.


packages/iframe-coordinator/src/messages/LabeledMsg.ts
```diff
@@ -9,12 +9,8 @@ import {
...
 } from 'decoders';
 
-// using `import` here with TypeScripts' requireJsonModule
-// option breaks TS' ability to generate type files, so we
-// fall back to `require`.
-
-// tslint:disable-next-line:no-var-requires
-const version = require('../../package.json').version;
+// This is replaced by rollup-plugin-replace with the actual version from package.json
+const version = '__PACKAGE_VERSION__';
```
MattCheely, line 13: The json import in the build was creating all kinds of problems with TypeScript and rollup, either failing completely or leading to weird syntax errors in other parts of the project that are perfectly valid. Rather than fight with it, I opted to just use a rollup plugin to replace this with the version. That's less code in the bundle anyway.

thomas-c-dillon:
Not a rollup expert but it looks good to me. It looks like we would be safe to remove the `webpack.config.cjs` file as well, yeah?

katie-bobbe-genesys (approved):

PR #216, opened 2022-11-23
communicate click events from client to host

thomas-c-dillon:


packages/iframe-coordinator/src/client.ts
```diff
@@ -355,34 +363,25 @@ bad input into one of the iframe-coordinator client methods.
...
+   * Turns off the behavior of intercepting link clicks in the client window click handler.
    */
   public stopInterceptingLinks(): void {
-    if (!this._isInterceptingLinksStarted) {
-      return;
-    }
-
-    this._isInterceptingLinksStarted = false;
-    this._clientWindow.removeEventListener('click', this._onWindowClick);
+    this._shouldInterceptLinks = true;
```
jordanstith15, line 384: Should this be set to false? Otherwise, both this and the startInterceptingLinks function do the same thing, which makes me wonder if there is a point to them at all

jordanstith15 (changes requested):

packages/iframe-coordinator/src/client.ts
thomas-c-dillon: Good catch!

jordanstith15 (approved):

packages/iframe-coordinator/src/messages/Click.ts
```diff
@@ -0,0 +1,18 @@
+import { constant, Decoder, object } from 'decoders';
+import { labeledDecoder, LabeledMsg } from './LabeledMsg';
+
+/**
+ * A message used to notify the host that a click (mousedown followed by mouseup)
+ * occurred in the client application.
+ */
+export interface LabeledClick extends LabeledMsg<'clickFired', {}> {
+  /** Message identifier */
+  msgType: 'clickFired';
```
MattCheely, line 10: Is it just me or is this name a bit ambiguous about if some active code is "firing" a click or if the user just clicked on something?  The terminology around events being fired/triggered/emitted is frustratingly varied. Would "appClicked" or "userClick" be clearer? It might just be me.

MattCheely (changes requested):
This all looks very good, however, I think we can make this work seamlessly for the host application by automatically listening for this event on the host side and calling [`click()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click) on the `frame-router` element when it's detected.

packages/iframe-coordinator/src/messages/Click.ts
thomas-c-dillon: I opted for consistency with the existing registeredKeyFired for KeyDown. I am happy to go with whatever name makes the most sense to everyone.

MattCheely (approved):

PR #226, opened 2023-02-10
feat(prompt-on-leave) - addition of prompt on leave

gavin-everett-genesys:
**Related Ticket :** https://inindca.atlassian.net/browse/COMUI-1289

### DIFF
--- a/packages/iframe-coordinator/src/messages/PromptOnLeave.ts
+++ b/packages/iframe-coordinator/src/messages/PromptOnLeave.ts
@@ -1,0 +1,36 @@
+import { boolean, constant, Decoder, object, optional, string } from 'decoders';
+import { labeledDecoder, LabeledMsg } from './LabeledMsg';
+
+/** A prompt on leave dialog to be displayed
+ * by the host application. If the host application receives a message with the shouldPrompt field set to true
+ * a dialog will be displayed asking the user for confirmation before navigating.
+ */
+export interface PromptOnLeave {
+  /** The host application will ask the user for confirmation before
+   * leaving the current page if it has received a message with the shouldPrompt field set to true.
+   */
+  shouldPrompt: boolean;
+  /** Optional message to prompt the user with. */
+  message?: string;
+}
+
+/**
+ * A message used to request a prompt on leave dialog to be displayed in the host app.
+ */
+export interface LabeledPrompt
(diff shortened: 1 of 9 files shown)


packages/iframe-coordinator/src/messages/PromptOnLeave.ts
```diff
@@ -0,0 +1,30 @@
+import { boolean, constant, Decoder, object } from 'decoders';
+import { labeledDecoder, LabeledMsg } from './LabeledMsg';
+
+/** A prompt on leave notification to be displayed
+ * by the host application
+ */
+export interface PromptOnLeave {
+  /** A payload of type boolean */
+  payload: boolean;
```
MattCheely, line 9: I would call this `shouldPrompt` or something else descriptive and update the doc comment to indicate that the host app will ask the user for confirmation before leaving the current page if it has received a message with the field set to true. 

I know I didn't capture this in the ticket, but after some research on the future of navigation APIs and what might be possible for us in the app, I'd also add an optional `message?: string` field documented as the message to prompt the user with (if possible).

packages/iframe-coordinator/src/messages/PromptOnLeave.ts
```diff
@@ -0,0 +1,30 @@
+import { boolean, constant, Decoder, object } from 'decoders';
+import { labeledDecoder, LabeledMsg } from './LabeledMsg';
+
+/** A prompt on leave notification to be displayed
+ * by the host application
```
MattCheely, line 5: I'd add some more detail to this to clarify how the message is intended to impact navigation. These comments are used in the [generated documentation](https://apps.inindca.com/common-ui-docs/iframe-coordinator/interfaces/client.NavRequest.html)

MattCheely (changes requested):
We should also update [client-setup.md](packages/iframe-coordinator/doc/client-setup.md) with a description of how to use `promptOnLeave` in the "Requesting Host Actions" section.  That will show up [here](https://apps.inindca.com/common-ui-docs/iframe-coordinator/modules/client.html) in the deployed docs.

MattCheely (changes requested):
Sorry for not catching it in the first round of reviews, but we should add some basic tests in `ClientToHost.spec.ts` for validation of the new message type.

jordanstith15 (approved):

PR #229, opened 2023-04-17
feat(frame-router): emit clientNotFound event after each invalid rout…

thomas-c-dillon:
…e change

COMUI-1632

### DIFF
--- a/packages/iframe-coordinator/src/elements/frame-router.ts
+++ b/packages/iframe-coordinator/src/elements/frame-router.ts
@@ -129,6 +129,15 @@ export default class FrameRouterElement extends HTMLElement {
         );
       }
 
+      if (clientInfo === null) {
+        /**
+         * Emit a clientNotFound event when there is not matching client.
+         * As legacy behavior, a clientChanged event will also fire unless the
+         * _currentClientId was already an empty string.
+         */
+        this.dispatchEvent(new CustomEvent('clientNotFound'));
+      }
+
       this._currentClientId = newClientId;
 
       const newLocation = this._frameManager.setFrameLocation(
--- a/packages/iframe-coordinator/src/elements/frame-router.ts
+++ b/packages/iframe-coordinator/src/elements/frame-router.ts
@@ -131,8 +131,8 @@ export default class FrameRouterElement extends HTMLElement {
 
       if (clientInfo === null) {
         /**
-         * Emit a clientNotFound event when there is not matching client.
-         * As legacy behavior, a clientChanged event will also fire unless the
+         * Emit a clientNotFound event when there is no matching client.
+         * As a legacy behavior, a clientChanged event will also fire unless the
          * _currentClientId was already an empty string.
          */
         this.dispatchEvent(new CustomEvent('clientNotFound'));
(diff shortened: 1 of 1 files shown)


jordanstith15 (approved):

packages/iframe-coordinator/src/elements/frame-router.ts
```diff
@@ -129,6 +129,15 @@ export default class FrameRouterElement extends HTMLElement {
         );
       }
 
+      if (clientInfo === null) {
+        /**
+         * Emit a clientNotFound event when there is not matching client.
```
gavin-everett-genesys, line 134: Minor but I would add an `a` after the `not`

packages/iframe-coordinator/src/elements/frame-router.ts
```diff
@@ -129,6 +129,15 @@ export default class FrameRouterElement extends HTMLElement {
...
+         * As legacy behavior, a clientChanged event will also fire unless the
```
gavin-everett-genesys, line 135: Should there be an `a` after the `As` here also.

gavin-everett-genesys (approved):

katie-bobbe-genesys:
I tried using this branch in the nav system project, and found that the event was always being triggered twice. Once after `changeRoute` is called from `attributeChangedCallback` and again when `changeRoute` is being called from `setupFrames`. Not sure if this is an issue or not.

thomas-c-dillon:
> I tried using this branch in the nav system project, and found that the event was always being triggered twice. Once after `changeRoute` is called from `attributeChangedCallback` and again when `changeRoute` is being called from `setupFrames`. Not sure if this is an issue or not.

I believe this is related to why this event is needed in the first place. I think Elm adds the clientConfig before the route, so when the clientConfig is supplied there is no route yet and therefore no matching client, then the route is added which also has no matching client and triggers a second event. 

How does it work on a valid route? We might get an initial clientNotFound and then a clientChanged immediately after. Hopefully, this doesn't cause a problem. This might also be related to the current behavior that is requiring a timeout in our 404 logic.

Perhaps we need a mechanism to queue an async changeRoute after a small delay when clientConfig or route is updated so we only get one routeChange per set of attribute changes. Or some other similar approach. @MattCheely Do you have any thoughts on this?

thomas-c-dillon:
With more local testing with logging I have confirmed the following:

- In the nav-system any time the url is updated the `changeRoute` method is called twice. Once for the clientConfig property and once for the route attribute.
- The ordering of these two code paths are not consistent. It is possible for the route attribute update to fire first (and the router to not exist).
- The only reason that we are not getting two `clientChanged` events is due to the check against `this._currentClientId`.

Possible solutions I have considered so far:
- Track `currentPath` along side `currentClientId` and do a similar check before emitting the event (the method continues to be called twice but we only emit one event).
- Refactor the existing code to check the clientConfig against the current clientConfig and only reconfigure the clients if there is a change. I think we still end up with two events on initial render though. Maybe additional logic can address this as well but this seems the most complex.

thomas-c-dillon:
Discussed with Matt and decided on just checking for the currentPath. This should get rid of the duplicate event.

katie-bobbe-genesys (approved):

PR #244, opened 2023-12-13
build(dependencies): updated root and app dependencies to latest

daragh-king-genesys:
updated root and app dependencies to latest

COMUI-2545


thomas-c-dillon (approved):
Ran the example apps and it seems to work as expected.

jordanstith15 (approved):

PR #250, opened 2024-01-08
build(dependencies): updated dependencies to wanted

daragh-king-genesys:
updated dependencies to wanted

COMUI-2541


packages/iframe-coordinator/package.json
```diff
@@ -19,21 +19,20 @@
...
+    "puppeteer": "^21.7.0",
+    "rollup": "^4.9.4",
+    "typedoc": "^0.25.7",
     "typescript": "^4.9.5"
   },
   "scripts": {
     "build": "npm run build-lib",
     "build-lib": "rollup -c",
-    "current-version": "cross-var \"echo $npm_package_version\"",
+    "current-version": "echo $npm_package_version",
```
daragh-king-genesys, line 35: I removed `cross-var` as it is an unmaintained package that has a dependency with a vulnerability. It is also overkill for an echo command that is basically already cross platform.

MattCheely (approved):

PR #252, opened 2024-01-09
test(karma): switch to jsdom test runner

MattCheely:
Chrome Headless is failing in CI since the puppeteer upgrade. Switch to JSDOM to simplify CI testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "iframe-coordinator-monorepo",
-  "version": "5.5.4",
+  "version": "5.5.6",
   "workspaces": [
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "iframe-coordinator-monorepo",
-  "version": "5.5.4",
+  "version": "5.5.6",
```
MattCheely, line 3: The package version also got out of sync with what's published somewhere along the way, so I bumped it in the repo to clear that out.

daragh-king-genesys (approved):

PR #270, opened 2024-06-24
fix: allowing routes with more specificity

jordanstith15:


apps/ifc-example-client/ifc-cli.config.js
```diff
@@ -16,6 +16,15 @@ module.exports = function (frameRouter) {
...
       },
+      application3: {
+        // We are removing the built-in proxy route example for now, as there are conflicts with vite and
+        // we will be adding improvements to the proxy behavior in the near future
+        url: `http://${hostname}:8080/client-app-2/#/`,
+        assignedRoute: "/app2/test",
+        allow: "camera http://localhost:8080;", // optional
+        sandbox: "allow-presentation allow-modals", // optional
+        defaultTitle: "iframe Application 2 Example", // optional, but needed for accessibility
+      },
```
jordanstith15, line 27: Added extra app to test the client id and routes with more specificity

packages/iframe-coordinator-cli/src/views/IframeEmbed.vue
```diff
@@ -5,6 +5,8 @@
       <span class="app-route">{{ frameRoute }}</span>
       as
       <span class="frame-url">{{ frameUrl }}</span>
+      with Client Id of
+      <span class="client-id">{{ currentClientId }}</span>
```
jordanstith15, line 9: showing current client id so it's easier to see and test whether the correct client is loaded

packages/iframe-coordinator/src/HostRouter.ts
```diff
@@ -13,9 +13,14 @@ export class HostRouter {
   private _clients: ClientInfo[];
 
   constructor(clients: RoutingMap) {
-    this._clients = Object.keys(clients).map((id) => {
-      return parseRegistration(id, clients[id]);
-    });
+    this._clients = Object.entries(clients)
```
jordanstith15, line 16: Noticed that we were doing essentially Object.entries and doing brackets to get array data, which I've seen flagged before on code analyzers.

packages/iframe-coordinator/src/HostRouter.ts
```diff
@@ -13,9 +13,14 @@ export class HostRouter {
...
+      .map(([id, data]) => {
+        return parseRegistration(id, data);
+      })
+      .sort(
+        (a, b) =>
+          b.assignedRoute.split("/").length - a.assignedRoute.split("/").length,
```
jordanstith15, line 22: Sorts clients by number of url segments descending

packages/iframe-coordinator/src/HostRouter.ts
```diff
@@ -26,7 +31,7 @@ export class HostRouter {
   public getClientTarget(route: string): ClientTarget | null {
     let clientTarget: ClientTarget | null = null;
 
-    this._clients.forEach((client) => {
+    for (let client of this._clients) {
```
jordanstith15, line 34: Noticed this issue as well.
It was continuing to loop after finding a matching client, causing it to seem like it was searching back to front.
Now it breaks the loop after finding a matching client and returns the first match.

packages/iframe-coordinator/src/HostRouter.ts
jordanstith15: No clue where the comma came from. I think the linter added it on commit

MattCheely (changes requested):
We should add a new unit test against `getClientTarget` in `HostRoute.spec.ts` that would have failed previously but now succeeds.

MattCheely:
I think this has to be flagged as a breaking change because it can change which app shows in response to a specific URL. Technically it was a bug, but in theory there could be an app out there that accidentally relies on this behavior.

MattCheely (approved):

PR #271, opened 2024-07-09
fix: refactoring hash logic to fix possible issue with query param on host route (i.e. timemachine override)

jordanstith15:
Closes [COMUI-1386](https://inindca.atlassian.net/browse/COMUI-1386)

[COMUI-1386]: https://inindca.atlassian.net/browse/COMUI-1386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/iframe-coordinator/src/elements/frame-router.ts
```diff
@@ -273,6 +273,6 @@ export default class FrameRouterElement extends HTMLElement {
       return hostUrlObject.href;
     }
     const trimmedUrl = stripTrailingSlash(hostUrl);
-    return window.location.hash ? `${trimmedUrl}/#` : trimmedUrl;
+    return window.location.hash ? `${trimmedUrl}#` : trimmedUrl;
```
MattCheely, line 276: I think the change as is has too much potential to break existing behavior that some apps could depend on. What we could probably get away with is preserving the existing behavior in most cases, but avoiding issues with query parameters by manipulating the hostUrlObject's `path` and `hash` properties to enforce that the path ends in `/` and the `hash` is set as `#` if not present. 

We should also add a story for the next major release to determine if we want to preserve this behavior at all.

MattCheely (changes requested):

MattCheely:
We should add a unit test that calls `setupFrames` on a `FrameRouter` instance and validates that the `hostRootUrl` property in the generated environment data is constructed as expected, both with and without query parameters in the original URL.

MattCheely (approved):

PR #280, opened 2025-01-09
Support building web-workers in the example app

MattCheely:
In order to enable the web-worker builds, ifc-cli was updated to be able to serve static files. This should also have some utility for users in being able to test/use pre-built client applications in ifc-cli.


apps/ifc-example-client/client-app.js
```diff
@@ -1,7 +1,3 @@
-// Load polyfills required for IE11
-import "@babel/polyfill";
-import "custom-event-polyfill";
-import "url-polyfill";
```
MattCheely, line 4: I also removed these old IE11 polyfills while I was working in these files

katie-bobbe-genesys (approved):

thomas-c-dillon (approved):

PR #281, opened 2025-01-09
fix the commit message on the license update

thomas-c-dillon:
I think this works?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2024 Genesys Cloud Services, Inc.
+Copyright (c) 2025 Genesys Cloud Services, Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


thomas-c-dillon:
Hmm, this seems to be working locally. Maybe the actions are testing main and not the branch being merged in?

packages/iframe-coordinator-cli/package.json
```diff
@@ -10,7 +10,7 @@
     "express": "^4.21.1",
     "find-root": "^1.1.0",
     "http-proxy-middleware": "^3.0.3",
-    "iframe-coordinator": "6.0.0",
```
katie-bobbe-genesys, line 13: Could we remove this version bump from this PR? I think we should only have the commitlint fix in this PR

katie-bobbe-genesys:
> Hmm, this seems to be working locally. Maybe the actions are testing main and not the branch being merged in?

Looks like it is testing main for the commitlint 
```
Run npm run lint

> iframe-coordinator-monorepo@6.1.0 lint
```
I think we should make a ticket to update the lint script so that it checks the current branch

packages/iframe-coordinator-cli/package.json
thomas-c-dillon: Sure, I can move that to a new branch. It will make the actions unable to `npm ci` and fail earlier. But it isn't like they are working anyway on this PR.

daragh-king-genesys (approved):

PR #286, opened 2025-01-23
fix: merge query for no-hash routes

cadgerfeast:
Hello there :)

I wanted to propose a fix for a particular use-case with client routes.

What I found is that when using hash router, everything works fine when specifying query in `route` attribute, though if the client has standard routing, then query params are encoded in pathname, breaking the iframe url.

The proposal here would be to merge query params both from host, and specified route, and then assign `pathname` and `search` with the computed values.

I also added two tests but would be happy to discuss about it if you feel that should not be a supported use-case.


MattCheely:
@cadgerfeast Can you clarify the use case for this and maybe provide some examples of current behavior vs expected behavior in specific scenarios? It would also be good to know if we have real-world use cases this is impacting. We can work through it here, or you're welcome to put something on my calendar or reach out in chat since we're both at Genesys.

PR #287, opened 2025-01-24
Add frame router tests

MattCheely:
In looking at https://github.com/purecloudlabs/iframe-coordinator/pull/271, I discovered some corner cases where behavior was changing. I was also able to get unit testing working for the underlying behavior, so I've opened the PR to add the unit tests and verify they are correct for the current release. This will capture the (admittedly inconsistent) current behavior of the library.  We can merge this and use it to refine the work in the query fix PR to ensure no regressions.


packages/iframe-coordinator/src/specs/FrameRouter.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    it("Adds a fragment to the tracked host URL if one is present on window but not on the provided root URL", () => {
+      window.location.hash = "foo";
+      const router = new FrameRouterElement();
+      router.clientConfig = {
+        clients: {},
+        envData: ENV_DATA,
+      };
+      //@ts-ignore
+      expect(router._envData).toEqual({
```
jordanstith15, line 49: "_envData" is accessible here? I thought it was a private variable

MattCheely: In TypeScript private is only meaningful within the type system. Because TS has to compile to vanilla JS, which has no direct concept of `private`, it just becomes a regular property on the object.  That means it's accessible in tests by disabling the type checks for that line. 

It's not a pattern I'd generally recommend though, and on the next major release when we're doing significant refactors, we should look for ways to make this testable without having to ignore type checks.

MattCheely:
Looks like I accidentally updated package-lock.json to point at our internal npm repo. I'll fix that and update the PR today.

jordanstith15 (approved):

PR #293, opened 2025-05-20
build(iframe-coordinator-cli): fix failing build due to package-lock.json mismatch

MattCheely:
The lockfile was out of sync, causing build failures. This was partly due to an unnecessary explicit dependency between the CLI package and a published version of the IFC library. NPM's automatic internal linking of workspace projects, combined with the vite bundling for ifc-cli makes that dependency unnecessary.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,12 +3,11 @@ import com.genesys.jenkins.Service
 
 def notifications = null
 String[] mailingList = [
-  "Matthew.Cheely@genesys.com",
-  "Daragh.King@genesys.com"
+  'CoreUI@genesys.com'
 ]
 
 def isReleaseBranch() {
-    return env.SHORT_BRANCH.equals('main') || env.SHORT_BRANCH.equals('master');
+  return env.SHORT_BRANCH.equals('main') || env.SHORT_BRANCH.equals('master')
 }
 
 pipeline {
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -3,12 +3,11 @@ import com.genesys.jenkins.Service
 
 def notifications = null
 String[] mailingList = [
-  "Matthew.Cheely@genesys.com",
-  "Daragh.King@genesys.com"
+  'CoreUI@genesys.com'
```
MattCheely, line 6: Apart from this change, everything else her is auto-formatting fixes

packages/iframe-coordinator-cli/package.json
```diff
@@ -10,7 +10,6 @@
     "express": "^4.21.1",
     "find-root": "^1.1.0",
     "http-proxy-middleware": "^3.0.3",
-    "iframe-coordinator": "6.2.0",
```
MattCheely, line 13: This dependency is not needed because NPM/vite will automatically look up imports of `iframe-coordinator` from the other packages in the workspace. The Vite build then bundles all of the IFC code into a single js file in `dist/`, such that it's also not a required dependency when installing ifc-cli in other projects.

thomas-c-dillon (approved):

daragh-king-genesys (approved):

PR #296, opened 2025-08-07
fix(frame-router): fix double-slash in host URL processing when path …

MattCheely:
…is `/`

The way we process host URLs before passing to the client triggers a corner case where the native URL type silently ignores certain changes to the `pathname` field.  This results in a bug where a path of `/` is represented to client apps as `//`.

COMUI-4108

### DIFF
--- a/packages/iframe-coordinator/src/elements/frame-router.ts
+++ b/packages/iframe-coordinator/src/elements/frame-router.ts
@@ -272,7 +272,9 @@ export default class FrameRouterElement extends HTMLElement {
     if (!hostUrlObject.hash) {
       hostUrlObject.pathname = stripTrailingSlash(hostUrlObject.pathname);
       if (window.location.hash) {
-        hostUrlObject.pathname += "/";
+        if (hostUrlObject.pathname !== "/") {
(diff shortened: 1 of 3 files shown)


packages/iframe-coordinator/src/elements/frame-router.ts
```diff
@@ -272,7 +272,9 @@ export default class FrameRouterElement extends HTMLElement {
     if (!hostUrlObject.hash) {
       hostUrlObject.pathname = stripTrailingSlash(hostUrlObject.pathname);
       if (window.location.hash) {
-        hostUrlObject.pathname += "/";
+        if (hostUrlObject.pathname !== "/") {
+          hostUrlObject.pathname += "/";
+        }
```
MattCheely, line 277: The previous code assumed we needed to add a `/` because we would always strip a `/` on line 273, however if the original `pathname` is `/`, the URL object will not change it to an empty string on line 273. Instead, it is preserved as `/`, in which case we should not re-add the `/`. 

We have future roadmap work planned to remove this sort of processing entirely, but it would be too disruptive for consuming apps to do right now.

PR #299, opened 2025-09-11
fix(ifc-cli.js): fix typo in proxy config file option configuration

heymagurany:
This PR fixes a typo in the configuration of the `--proxy-config` CLI option.


MattCheely:
I've moved the fix to an internal branch to fix the CI issues. We'll PR and merge that one.

